PR #47, opened 2021-08-31
Print messages for token auth scenarios

JuanitoFatas:
This PR prints information for users when:

- No API key was supplied
- Invalid API key supplied
- Unknown errors

### DIFF
--- a/lib/rspec/buildkite/analytics/uploader.rb
+++ b/lib/rspec/buildkite/analytics/uploader.rb
@@ -141,7 +141,7 @@ module RSpec::Buildkite::Analytics
               end
             else
               request_id = response.to_hash["x-request-id"]
-              puts "Unknown error. Please contact support with this request ID `#{request_id}` support+analytics@buildkite.com."
+              puts "Unknown error. If this error persists, please contact support+analytics@buildkite.com with this request ID `#{request_id}`."
             end
           else
             puts "No Suite API key provided. You can get the API key from your Suite settings page."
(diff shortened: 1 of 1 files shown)


jasmine-q (approved):

lib/rspec/buildkite/analytics/uploader.rb
```diff
@@ -130,13 +130,21 @@ def self.configure
...
+              puts "Invalid Suite API key. Please double check your Suite API key."
+            when "200", "201"
               json = JSON.parse(response.body)
 
               if (socket_url = json["cable"]) && (channel = json["channel"])
                 RSpec::Buildkite::Analytics.session = Session.new(socket_url, authorization_header, channel)
               end
+            else
+              request_id = response.to_hash["x-request-id"]
+              puts "Unknown error. Please contact support with this request ID `#{request_id}` support+analytics@buildkite.com."
```
blaknite, line 144:
```suggestion
              puts "Unknown error. If this error persists, please contact support+analytics@buildkite.com with this request ID `#{request_id}`."
```

lib/rspec/buildkite/analytics/uploader.rb
```diff
@@ -130,13 +130,21 @@ def self.configure
               puts "Error communicating with the server: #{e.message}"
             end
 
-            if response.is_a?(Net::HTTPSuccess)
+            case response.code
+            when "401"
+              puts "Invalid Suite API key. Please double check your Suite API key."
+            when "200", "201"
```
niceking, line 136: Just wondering why we're accounting for both a 200 and a 201 return here? Like does the server not consistently return the same status code for this initial auth request?? And if it doesn't, should we be accounting for more values in the 2XX range then?

JuanitoFatas: I was trying to be generous, but we always return 200. Fixed in c18fdf21cb6350fa809ca40749f1e97ecd7fec06. Thanks!

PR #49, opened 2021-09-06
Require one less env

JuanitoFatas:
We currently require two ENVs `BUILDKITE` and `BUILDKITE_BUILD_ID` to make Runs work (that Jobs inside a Build will be filed under one Run). This PR changes to only require `BUILDKITE_BUILD_ID`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,16 +1,16 @@
 PATH
   remote: .
   specs:
-    rspec-buildkite-analytics (0.2.0)
-      activesupport
-      rspec-core
-      rspec-expectations
-      websocket
+    rspec-buildkite-analytics (0.3.2)
+      activesupport (>= 5.2, <= 7.0)
+      rspec-core (~> 3.10)
+      rspec-expectations (~> 3.10)
+      websocket (~> 1.2)
 
 GEM
   remote: https://rubygems.org/
   specs:
-    activesupport (6.1.4)
+    activesupport (6.1.4.1)
       concurrent-ruby (~> 1.0, >= 1.0.2)
       i18n (>= 1.6, < 2)
       minitest (>= 5.1)
--- a/lib/rspec/buildkite/analytics/ci.rb
+++ b/lib/rspec/buildkite/analytics/ci.rb
@@ -4,22 +4,15 @@ require "securerandom"
 
 module RSpec::Buildkite::Analytics::CI
   def self.env
-    if ENV["BUILDKITE"]
-      {
-        "CI" => "buildkite",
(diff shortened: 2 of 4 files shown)


Gemfile.lock
```diff
@@ -1,16 +1,16 @@
...
+      activesupport (>= 5.2, <= 7.0)
+      rspec-core (~> 3.10)
+      rspec-expectations (~> 3.10)
+      websocket (~> 1.2)
 
 GEM
   remote: https://rubygems.org/
   specs:
-    activesupport (6.1.4)
+    activesupport (6.1.4.1)
```
jasmine-q, line 13: were these meant to change?

JuanitoFatas: yup `Gemfile.lock` change is fine for our gem development. It does not affect users of this gem. It is also possible to have gem without `Gemfile.lock`.

spec/analytics/ci_spec.rb
```diff
@@ -23,7 +23,13 @@
 
       expect(result).to match({
         "CI" => nil,
-        "key" => "845ac829-2ab3-4bbb-9e24-3529755a6d37"
```
niceking, line 26: I think on line 19 above this should probably change to something like "BUILDKITE env variable not set"

niceking (approved):

jasmine-q (approved):

lib/rspec/buildkite/analytics/ci.rb
```diff
@@ -4,22 +4,15 @@
...
-        "job_id" => ENV["BUILDKITE_JOB_ID"],
-        "message" => ENV["BUILDKITE_MESSAGE"]
-      }
-    else
-      {
-        "CI" => nil,
-        "key" => SecureRandom.uuid
-      }
-    end
+    env = {
```
blaknite, line 7:
```ruby
      if ENV["BUILDKITE_BUILD_ID"]
       {
         "CI" => "buildkite",
         "key" => ENV["BUILDKITE_BUILD_ID"],
         "url" => ENV["BUILDKITE_BUILD_URL"],
         "branch" => ENV["BUILDKITE_BRANCH"],
         "commit_sha" => ENV["BUILDKITE_COMMIT"],
         "number" => ENV["BUILDKITE_BUILD_NUMBER"],
         "job_id" => ENV["BUILDKITE_JOB_ID"],
         "message" => ENV["BUILDKITE_MESSAGE"]
       }
     else
       {
         "CI" => nil,
         "key" => SecureRandom.uuid
       }
     end
```

PR #50, opened 2021-09-07
Only take the commit message title

JuanitoFatas:
A Commit the subject and body usually separates by a newline. This PR changes the gem to only captures the subject.


JuanitoFatas:
change of plans

PR #51, opened 2021-09-08
Only use BUILDKITE_BUILD_ID but keep previous env behaviour

blaknite:
This PR restores the original behaviour of the CI check by always passing "buildkite" as the CI key. It keeps the new behaviour of only requiring and checking for BUILDKITE_BUILD_ID.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    rspec-buildkite-analytics (0.3.2)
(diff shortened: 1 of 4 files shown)


JuanitoFatas (approved):
LGTM 👍🏼 

Could you also bump the version in this PR and I can release a new version for you.

blaknite:
> Could you also bump the version in this PR and I can release a new version for you.

Done and thanks so much!

PR #54, opened 2021-10-19
Add a dependency version for openssl

niceking:
I made this `>=2.2.1` cos the bugfix is only in the releases `2.1.3` and `2.2.1` 
Screen Shot 2021-10-20 at 11 50 57 AM

I then bundled buildkite/buildkite locally with the gem and ran `BUILDKITE_ANALYTICS_TOKEN=local rspec spec/models` and all executions were created as expected, so I think this will be fine 🤞


jasmine-q (approved):
🤞

niceking:
Closing in favour of #55

PR #62, opened 2021-11-04
Add debug mode

niceking:
This PR introduces a debug mode to the collector, which creates a text file with debug information that is uploaded to the job as an artifact. This was done because we realised that it wasn't possible with Datadog to exhaustively log the requests going through the socket, and it made debugging errors difficult. Hence, saving the relevant info to a text file, and also [adding logging on Buildkite](https://github.com/buildkite/buildkite/pull/7365) so we can build a complete picture of what has/hasn't been received. 

I have added two new environment params, `BUILDKITE_ANALYTICS_DEBUG_ENABLED` and `BUILDKITE_ANALYTICS_DEBUG_FILEPATH` which configure the enabling of debug mode, and the filepath to save the artifact at. I created a custom lightweight Logger class to facilitate the logging, it needed to be threadsafe so it's using a `Queue` object to store the log lines. 

Another change in this PR is that I've changed the payload of the `end_of_transmission` message to include some counts of how many examples were run, and how many failed. This is so that in the future we can compare these numbers against the number of executions that were actually created in each upload. Because there's going to be a lot of log files, I wanted there to be some mechanism where we could winnow down to log files that we're interested in, rather than just hoping that there's interesting info. 

To achieve the above, I had to move some of the code in the `after(:suite)` hook into a formatter instead, as the example counts I wanted weren't available in the after suite callback. 

I tested this PR by bundling locally with the changes in [this PR](https://github.com/buildkite/buildkite/pull/7365)

### DIFF
--- a/lib/rspec/buildkite/analytics.rb
+++ b/lib/rspec/buildkite/analytics.rb
@@ -15,11 +15,15 @@ module RSpec::Buildkite::Analytics
     attr_accessor :url
     attr_accessor :uploader
     attr_accessor :session
+    attr_accessor :debug_enabled
+    attr_accessor :debug_filepath
   end
 
-  def self.configure(token: nil, url: nil)
+  def self.configure(token: nil, url: nil, debug_enabled: false, debug_filepath: nil)
     self.api_token = token || ENV["BUILDKITE_ANALYTICS_TOKEN"]
     self.url = url || DEFAULT_URL
+    self.debug_enabled = debug_enabled || !!(ENV["BUILDKITE_ANALYTICS_DEBUG_ENABLED"])
+    self.debug_filepath = debug_filepath || ENV["BUILDKITE_ANALYTICS_DEBUG_FILEPATH"]
 
     require_relative "analytics/uploader"
 
--- a/lib/rspec/buildkite/analytics/reporter.rb
+++ b/lib/rspec/buildkite/analytics/reporter.rb
@@ -1,6 +1,6 @@
 module RSpec::Buildkite::Analytics
   class Reporter
-    RSpec::Core::Formatters.register self, :example_passed, :example_failed, :example_pending
+    RSpec::Core::Formatters.register self, :example_passed, :example_failed, :example_pending, :dump_summary
 
     attr_reader :output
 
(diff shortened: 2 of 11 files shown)


niceking:
[bk-analytics-2021-11-02-20:42:30-.log](https://github.com/buildkite/rspec-buildkite-analytics/files/7471430/bk-analytics-2021-11-02-20.42.30-.log)

niceking:
Example log file 
[bk-analytics-2021-11-02-20:42:30-.log](https://github.com/buildkite/rspec-buildkite-analytics/files/7471431/bk-analytics-2021-11-02-20.42.30-.log)

lib/rspec/buildkite/analytics/session.rb
```diff
@@ -12,6 +12,28 @@ class Session
...
+        @log << "#{Time.now.strftime("%F-%R:%S.%9N")} #{Thread.current} #{str}"
+      end
+
+      def to_array
+        # This empty check is important cos calling pop on a Queue is blocking until
+        # it's not empty
+        if @log.empty?
+          []
+        else
+          Array.new(@log.size) { @log.pop }
```
JuanitoFatas, line 30: Would this only pops the last element from the queue instead of all elements from the queue? 🤔

niceking: If you pass a block to `Array.new` it populates it with the given object, and since I set the size of the Array to be the size of the queue, pop will be called the right number of times. There's further examples at the top of [this page](https://ruby-doc.org/core-2.7.0/Array.html)

I found this code snippet on [this stackoverflow post](https://stackoverflow.com/questions/34481288/ruby-queue-to-array), happy to choose a different way of doing it if you feel like it's unclear?

JuanitoFatas: Ah what I meant is suppose we have 2 errors in `@log`:

```ruby
@log << "error 1"
@log << "error 2"
```

This line will create any array of `["error 1", "error 1"]` instead of `["error 1", "error 2"]` right?

Is this what we want? 🤔

niceking: Ruby `Queue` is a FIFO queue and the pop method removes the first item in the queue, [docs here](https://ruby-doc.org/core-2.5.0/Queue.html#method-i-pop). Also I attached a log file so you can see that it's working :)

JuanitoFatas: The log file looks good! 👍🏼

blaknite (approved):

lib/rspec/buildkite/analytics.rb
```diff
@@ -15,11 +15,15 @@ class << self
...
     attr_accessor :session
+    attr_accessor :debug_enabled
+    attr_accessor :debug_filepath
   end
 
-  def self.configure(token: nil, url: nil)
+  def self.configure(token: nil, url: nil, debug_enabled: false, debug_filepath: nil)
     self.api_token = token || ENV["BUILDKITE_ANALYTICS_TOKEN"]
     self.url = url || DEFAULT_URL
+    self.debug_enabled = debug_enabled || (ENV["BUILDKITE_ANALYTICS_DEBUG_ENABLED"] == "true" ? true : false)
```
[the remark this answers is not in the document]
niceking: Love a @JuanitoFatas special 😍

lib/rspec/buildkite/analytics/reporter.rb
```diff
@@ -1,6 +1,6 @@
 module RSpec::Buildkite::Analytics
   class Reporter
-    RSpec::Core::Formatters.register self, :example_passed, :example_failed, :example_pending
+    RSpec::Core::Formatters.register self, :example_passed, :example_failed, :example_pending, :dump_summary
```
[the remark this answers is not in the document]
niceking: Sad that I can't poo react to your poo

PR #63, opened 2021-11-07
Avoid activesupport when generating debugging timestamps

yob:
I believe `DateTime.current` is an activesupport-ism.

This gem does depend on activesupport so in theory we could use `DateTime.current`, however when I was running v0.4.0 with my non-rails rspec project I was getting:

```ruby
NameError: uninitialized constant RSpec::Buildkite::Analytics::Reporter::DateTime
Did you mean?  DateAndTime
  /.../rspec-buildkite-analytics/lib/rspec/buildkite/analytics/reporter.rb:36:in `dump_summary'
```

I think we could fix this by adding an explicit `require "datetime"` to the file. However if we're going to do that, we may as well stick to the stdlib and use plain `Time.now` when generating a timestamp.

### DIFF
--- a/lib/rspec/buildkite/analytics/reporter.rb
+++ b/lib/rspec/buildkite/analytics/reporter.rb
@@ -1,3 +1,5 @@
+require "time"
+
 module RSpec::Buildkite::Analytics
   class Reporter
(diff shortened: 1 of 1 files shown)


niceking (approved):
Oh nice, thanks!

PR #65, opened 2021-11-10
Move send to background thread

niceking:
I've added another thread to this gem but I swear it hasn't gotten more complicated 🙈

The major change here is that there's a new send queue for outbound messages. When an rspec test is finished, the results from that test are pushed into the send queue, rather than being written to the socket immediately. Instead, a new `write_thread` is instantiated that waits for messages to appear on the send queue, and sends them through the socket. This change means that reading and writing from the socket happen entirely in the two background threads, so any failures in them don't affect the main thread which is running the rspec tests. If the socket connection is broken for any reason, the existing reconnection + retransmission logic handles this. 

When the rspec tests have finished, our existing `session.close` behaviour was waiting up to 75 seconds for the last `confirm` message to be sent by the server. Since sending through the socket is really fast, any build up of messages (because of disconnects) can catch up in this time and then we can finish and close out the session as normal. Added some logic around only queueing one `eot` message at a time, as we don't know when things are sent anymore. 

Unit testing for this has gotten a little more complicated due to the extra threads, so I haven't really made many modifications to this. Instead I manually tested these cases: 
- works when an initial connection to buildkite can't be established (no executions + s3 data in this scenario)
- works and all executions + s3 data is persisted when connection to buildkite can be established (tested both for a small test run, and one that spans multiple upload parts)
- works and all executions + s3 data is persisted when a reconnection to buildkite is successful 
- doesn't explode if reconnection to buildkite is unsuccessful (doesn't affect the rspec test running, but it does wait for full length of timeout before closing the session)

~~Also found [this bug](https://github.com/buildkite/buildkite/pull/7454) while testing~~

### DIFF
--- a/lib/rspec/buildkite/analytics/session.rb
+++ b/lib/rspec/buildkite/analytics/session.rb
@@ -40,6 +40,7 @@ module RSpec::Buildkite::Analytics
 
       @unconfirmed_idents = {}
       @idents_mutex = Mutex.new
+      @send_queue = Queue.new
       @empty = ConditionVariable.new
       @closing = false
       @reconnection_mutex = Mutex.new
@@ -108,6 +109,8 @@ module RSpec::Buildkite::Analytics
 
       # Then we always disconnect cos we can't wait forever? 🤷‍♀️
       @connection.close
+      # We kill the write thread cos it's got a while loop in it, so it won't finish otherwise
+      @write_thread.kill
       @logger.write("socket connection closed")
     end
 
@@ -132,13 +135,9 @@ module RSpec::Buildkite::Analytics
     end
 
     def write_result(result)
-      result_as_json = result.as_json
+      queue_and_track_result(result.id, result.as_json)
 
-      add_unconfirmed_idents(result.id, result_as_json)
-
-      transmit_results([result_as_json])
-
-      @logger.write("transmitted #{result.id}")
+      @logger.write("added #{result.id} to send queue")
     end
 
     def unconfirmed_idents_count
@@ -148,17 +147,6 @@ module RSpec::Buildkite::Analytics
     end
 
     private
-
-    def transmit_results(results_as_json)
-      @connection.transmit({
-        "identifier" => @channel,
-        "command" => "message",
-        "data" => {
-          "action" => "record_results",
-          "results" => results_as_json
-          }.to_json
-        })
-    end
 
     def connect
       @logger.write("starting socket connection process")
@@ -177,6 +165,29 @@ module RSpec::Buildkite::Analytics
       wait_for_confirm
 
       @logger.write("connected")
+
+      @write_thread = Thread.new do
+        @logger.write("hello from write thread")
+        # Pretty sure this eternal loop is fine cos the call to queue.pop is blocking
+        loop do
+          data = @send_queue.pop
+          message_type = data["action"]
+          @connection.transmit({
+            "identifier" => @channel,
+            "command" => "message",
+            "data" => data.to_json
+          })
+
+          if RSpec::Buildkite::Analytics.debug_enabled
+            data = JSON.parse(message["data"])
+            message_type = data["action"]
+            ids = if message_type == "record_results"
+              data["results"].map { |result| result["id"] }
+            end
+            @logger.write("transmitted #{message_type} #{ids}")
+          end
+        end
+      end
     end
 
     def pop_with_timeout(message_type)
@@ -201,9 +212,16 @@ module RSpec::Buildkite::Analytics
       end
     end
 
-    def add_unconfirmed_idents(ident, data)
+    def queue_and_track_result(ident, result_as_json)
       @idents_mutex.synchronize do
-        @unconfirmed_idents[ident] = data
+        @unconfirmed_idents[ident] = result_as_json
+
+        data = {
+          "action" => "record_results",
+          "results" => [result_as_json]
+        }
+
+        @send_queue << data
       end
     end
 
@@ -228,17 +246,22 @@ module RSpec::Buildkite::Analytics
 
     def send_eot
       # Expect server to respond with data of indentifiers last upload part
+      data = {
+        "action" => "end_of_transmission",
+        "examples_count" => @examples_count.to_json
+      }
 
-      @connection.transmit({
-        "identifier" => @channel,
-        "command" => "message",
-        "data" => {
-          "action" => "end_of_transmission",
-          "examples_count" => @examples_count.to_json
-        }.to_json
-      })
+      # We grab the idents mutex here (even though we aren't modifying idents)
+      # because the idents array needs to stay in sync with messages
+      # that are put on the send queue. There just isn't any
+      # corresponding ident to track here, but we still grab the
+      # mutex in case there were any last results to be added
+      # to the queue
+      @idents_mutex.synchronize do
+        @send_queue << data
+      end
 
-      @logger.write("transmitted EOT")
+      @logger.write("added EOT to send queue")
     end
 
     def process_message(data)
@@ -254,9 +277,20 @@ module RSpec::Buildkite::Analytics
       end
     end
 
-    def retransmit
-      data = @idents_mutex.synchronize do
-        @unconfirmed_idents.values
+    def queue_retransmit_message
+      @idents_mutex.synchronize do
+        results = @unconfirmed_idents.values
+
+        # queue the contents of the buffer, unless it's empty
+        unless results.empty?
+          data = {
+            "action" => "record_results",
+            "results" => results.to_json
+          }
+
+          @logger.write("retransmitted results #{@unconfirmed_idents.keys}")
+          @send_queue << data
+        end
       end
 
       # send the contents of the buffer, unless it's empty
--- a/lib/rspec/buildkite/analytics/session.rb
+++ b/lib/rspec/buildkite/analytics/session.rb
@@ -35,7 +35,7 @@ module RSpec::Buildkite::Analytics
     attr_reader :logger
 
     def initialize(url, authorization_header, channel)
-      @queue = Queue.new
+      @establish_subscription_queue = Queue.new
       @channel = channel
 
       @unconfirmed_idents = {}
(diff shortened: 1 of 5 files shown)


lib/rspec/buildkite/analytics/session.rb
```diff
@@ -134,11 +142,9 @@ def handle(_connection, data)
     def write_result(result)
       result_as_json = result.as_json
```
blaknite, line 143: I think I've commended on how the `as_json` method is mighty confusingly named. IIRC it's because we didn't want to overload to_json or to_h? It's doubly confusing because it's not returning json at all but a hash.

This makes all the places where `to_json` gets called in this PR seem odd as well. My brain is going "but this is already json, why are we calling `to_json` on it?

I think now might be a good time for a bit of a refactor/rename.

lib/rspec/buildkite/analytics/session.rb
```diff
@@ -177,11 +172,53 @@ def connect
...
+            end
+            @eot_queued_mutex.synchronize do
+              @eot_queued = false
+            end
+          end
+
+          @connection.transmit({
+            "identifier" => @channel,
+            "command" => "message",
+            "data" => data.to_json
```
blaknite, line 205: `results` in this value has already had `to_json` called on it on line 247. This results in a json object that gets reencoded as a json escaped string. Might this have something to do with your PR to bk/bk?

lib/rspec/buildkite/analytics/session.rb
```diff
@@ -177,11 +172,53 @@ def connect
...
+
+      # As this connect method can be called multiple times in the
+      # reconnection process, kill prev write threads (if any) before
+      # setting up the new one
+      @write_thread&.kill
+
+      @write_thread = Thread.new do
+        @logger.write("hello from write thread")
+        # Pretty sure this while loop is fine cos the call to queue.pop is blocking
+        while
```
blaknite, line 184: Instead of a `while` without condition we can use `loop`
```suggestion
        loop
```

lib/rspec/buildkite/analytics/session.rb
```diff
@@ -177,11 +172,53 @@ def connect
...
+          data = @send_queue.pop
+          message_type = data["action"]
+
+          if message_type == "end_of_transmission"
+            # Because of the unpredictable sequencing between the test suite finishing
+            # (EOT gets queued) and disconnections happening (retransmit results gets
+            # queued), we don't want to send an EOT before any retransmits are sent.
+            unless @send_queue.empty?
```
blaknite, line 192: For this I find `if any?` easier to follow.
```suggestion
            if @send_queue.any?
```

lib/rspec/buildkite/analytics/session.rb
```diff
@@ -177,11 +172,53 @@ def connect
...
+
+          @connection.transmit({
+            "identifier" => @channel,
+            "command" => "message",
+            "data" => data.to_json
+          })
+
+          if RSpec::Buildkite::Analytics.debug_enabled
+            ids = if message_type == "record_results"
+              results = JSON.parse(data["results"])
```
blaknite, line 210: I think the `JSON.parse` is another artefact of calling `to_json` twice.

lib/rspec/buildkite/analytics/session.rb
```diff
@@ -201,9 +238,16 @@ def wait_for_confirm
...
 
-    def add_unconfirmed_idents(ident, data)
+    def queue_and_track_result(ident, result_as_json)
       @idents_mutex.synchronize do
-        @unconfirmed_idents[ident] = data
+        @unconfirmed_idents[ident] = result_as_json
+
+        data = {
+          "action" => "record_results",
+          "results" => [result_as_json].to_json
```
blaknite, line 247: This `to_json` is the culprit I'm talking about above.

blaknite (commented):
This proved much easier to follow than I was expecting for a PR that adds more threads 🙌

I think [#7454](https://github.com/buildkite/buildkite/pull/7454) is an artefact of a bug in this PR which I've mentioned below.

yob:
I'm catching up after a long weekend and haven't had a chance to read this in detail yet.

but:

> Also found this bug while testing

This is a great find! Is it worth pulling the client-side part of this fix into a dedicated PR? Sounds like it'd be a small diff, and it's orthogonal to the work around threads and reliability. It'd be nice to have the bugfix in a dedicated PR with some commentary around it.

blaknite:
> This is a great find!

@yob I think it's a bug introduced by this PR due to double calling `to_json`.

lib/rspec/buildkite/analytics/session.rb
yob: should that be `loop do`?

yob:
Ooooh. Sorry for the noise 😬

yob (commented):
On threads:

I'm a bit nervous about having two threads operating on a single TCP socket, and wonder if the long term maintainability of the gem will be easier by moving all sending+receiving to a single background thread. This isn't a blocker for me, just some feedback.

I agree with @blaknite that in the context of this particular diff, adding the second thread hasn't made the code particularly more complex. My concern is more about reasoning about ordering and logic flow in the future, particularly when trying to track down bugs, etc.

On serialisation:

I haven't read this diff close enough to pick up the double-JSON issues that @blaknite has flagged, mainly because he put the time in first and maybe I don't need to 😅

I agree with @blaknite that the time required to follow the serialisation logic and what gets converted to JSON when is non-trivial. It's an existing characteristic of this file and not really made worse by this diff, but as a reviewer it'd be nice if the serialisation logic was easier to follow. Maybe a dedicated PORO for the results rather than arrays and strings and hashes and JSON-strings would help?

lib/rspec/buildkite/analytics/session.rb
niceking: Unfortunately this is a Queue and not an Enumerable so this method isn't available! https://ruby-doc.org/core-2.5.0/Queue.html

niceking:
@blaknite yeah great catch on the serialization, and I totally got confused by the misnomer of the `as_json` method 😅

@yob 

> I'm a bit nervous about having two threads operating on a single TCP socket
Yeah I hear that but also this has been part of the operation of this gem since its inception - previously it was the main thread and the read thread, and now it's a dedicated read thread and write thread. Other than the index error bug (which occurred less than 10 times in the time that Buildkite's had it installed) there hasn't really been any repercussions of this, and a refactor of this into one thread seems more risky tbh!

lib/rspec/buildkite/analytics/session.rb
```diff
@@ -132,13 +140,11 @@ def handle(_connection, data)
     end
 
     def write_result(result)
-      result_as_json = result.as_json
-
-      add_unconfirmed_idents(result.id, result_as_json)
+      result_as_hash = result.as_hash
```
blaknite, line 143:
```suggestion

```

lib/rspec/buildkite/analytics/session.rb
```diff
@@ -132,13 +140,11 @@ def handle(_connection, data)
...
 
-      transmit_results([result_as_json])
+      queue_and_track_result(result.id, result_as_hash)
```
blaknite, line 145:
```suggestion
      queue_and_track_result(result.id, result.as_hash)
```

lib/rspec/buildkite/analytics/session.rb
blaknite: what about?
```suggestion
            if @send_queue.length > 0
```

lib/rspec/buildkite/analytics/tracer.rb
```diff
@@ -13,15 +13,15 @@ def initialize(section, start_at, end_at, detail)
...
         {
           section: section,
           start_at: start_at,
           end_at: end_at,
           duration: end_at - start_at,
           detail: detail,
-          children: children.map(&:as_json),
-        }
+          children: children.map(&:as_hash),
+        }.with_indifferent_access
```
blaknite, line 24: Why `with_indifferent_access`?

blaknite (approved):
Renaming that method made this far easier to follow 🥳

Few questions and minor changes, otherwise 🚀

lib/rspec/buildkite/analytics/tracer.rb
niceking: Not sure exactly why this was happening, but the send queue data for a single result was a hash with symbol keys, but for all the results was a hash with string keys?? So [this line](https://github.com/buildkite/rspec-buildkite-analytics/pull/65/files#diff-e04810cfd3eb5dfc5d3661dbedcbf1ca7bfdc3d88e0c9a01d2ddab0fec126260R208) would fail as the ids couldn't always be accessed via the symbol attribute. Making the underlying structure a hash with indifferent access seemed like the easiest way of getting around it

niceking:
Had a convo with @blaknite about release process, and have decided that we'll merge this into main, and not do a gem release for it, but instead change buildkite's ref for this gem to git main. We'll leave in for a week or two and if nothing explodes, will then do a release then. Seems like the least faffing we could do (as compared to making a different branch to release this from) and it's pretty low risk that we'd need to do some urgent customer fix in the mean time

lib/rspec/buildkite/analytics/tracer.rb
JuanitoFatas: Active Support dependency got removed in https://github.com/buildkite/rspec-buildkite-analytics/pull/63 — So we need to do this in Ruby hmm

yob: #63 doesn't fully remove the AS dependency. It's still marked as a dep of the gem, and it uses activesupport notifications.

In #63 I was avoiding using AS methods though, because for some reason they weren't being added to `DateTime` in my non-rails app. Possibly because my app didn't require `date`?

JuanitoFatas: Ooops sorry I did not read the Pull Request in depth. We can still use `with_indifferent_access` 👍🏼

JuanitoFatas: > In #63 I was avoiding using AS methods though, because for some reason they weren't being added to `DateTime` in my non-rails app. Possibly because my app didn't require `date`?

[`DateTime.current`](https://api.rubyonrails.org/classes/DateTime.html#method-c-current) is a Rails extension so not available in non-Rails app.

PR #67, opened 2021-11-19
Add rubocop

niceking:
WIP, already found some bugs 😅


JuanitoFatas:
We will revisit this later

PR #70, opened 2021-12-08
Send failure reasons for failed executions

niceking:
This PR adds support for sending through two new pieces of data `failure_reason` and `failure_expanded`. There was a previous key called `failure_message` that I've removed as we never did anything with it. The content of this key was pretty much what is in `failure_reason`. 

`failure_reason` is a one line summary of the failure, e.g. `"Failure/Error: expect(true).to eq false"`. `failure_expanded` is an array, with each element being an array of `message_lines` and `backtrace` so that these elements can be formatted separately in the front end. 

I wasn't sure if it was worth adding a test for this without basically mocking out all of RSpec so I didn't 🤷‍♀️ I tested this with the three types of exception we'd expect, there's ones for failed expectation, multiple failed expectations and errors.

### DIFF
--- a/lib/rspec/buildkite/analytics/reporter.rb
+++ b/lib/rspec/buildkite/analytics/reporter.rb
@@ -18,6 +18,7 @@ module RSpec::Buildkite::Analytics
 
       if trace
         trace.example = example
+        trace.failure_reason, trace.failure_expanded = failure_info(notification) if example.execution_result.status == :failed
         RSpec::Buildkite::Analytics.session&.write_result(trace)
       end
     end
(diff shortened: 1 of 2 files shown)


lib/rspec/buildkite/analytics/reporter.rb
```diff
@@ -18,6 +18,7 @@ def handle_example(notification)
 
       if trace
         trace.example = example
+        populate_failure_info(trace, notification)
```
blaknite, line 21: I think we should move the mutation of the trace out of the `populate_failure_info` method. It could return an array of `[reason, expanded]` and then destructure like so.
```suggestion
        trace.failure_reason, trace.failure_expanded = failure_info(notification)
```

lib/rspec/buildkite/analytics/reporter.rb
```diff
@@ -18,6 +18,7 @@ def handle_example(notification)
 
       if trace
         trace.example = example
+        trace.failure_reason, trace.failure_expanded = failure_info(notification) if example.execution_result.status == :failed
```
blaknite, line 21: Shorter line length? Not overly fussed.

```suggestion
        if example.execution_result.status == :failed
          trace.failure_reason, trace.failure_expanded = failure_info(notification)
        end
```

blaknite (approved):
🚀

lib/rspec/buildkite/analytics/reporter.rb
niceking: 🤷‍♀️ I'll leave it for the rubocop PR

PR #74, opened 2022-01-13
Handle scenario when we got no response

JuanitoFatas:
An early return to avoid calling methods on no response code paths.


blaknite (approved):
🚀

PR #76, opened 2022-01-17
Add retries to initial connection

niceking:
Enough customers reported seeing the "Timeout: waited 30s for welcome" error during the initial connection process that I thought it'd be good to put retries around this, like we do in the disconnect. 

My first thought was to refactor the reconnection code to be shared, but realised that this wasn't going to work as the reconnection code needs a mutex around reconnection attempts because we're not sure which thread will be doing the reconnection. Instead, I refactored the `connect` method to remove the instantiation of the write thread, so the `connect` now is just doing the ActionCable handshake business. This also means that we don't need to worry about locking in the initial connection phase, as the write thread doesn't exists yet, and we have a smaller set of errors that we expect during initial connection. 

Have integration tested to see that reconnection + failed initial connection still works.

### DIFF
--- a/lib/rspec/buildkite/analytics/session.rb
+++ b/lib/rspec/buildkite/analytics/session.rb
@@ -58,9 +58,9 @@ module RSpec::Buildkite::Analytics
         reconnection_count += 1
         connect
       rescue TimeoutError, InitialConnectionFailure => e
-        @logger.write("failed initial connection attempt #{reconnection_count} due to #{e}")
+        @logger.write("rspec-buildkite-analytics could not establish an initial connection with Buildkite due to #{e}. Attempting retry #{reconnection_count} of #{MAX_RECONNECTION_ATTEMPTS}...")
         if reconnection_count > MAX_RECONNECTION_ATTEMPTS
-          $stderr.puts "rspec-buildkite-analytics could not establish an initial connection with Buildkite due to #{e.message}. You may be missing some data for this test suite, please contact support if this issue persists."
+          $stderr.puts "rspec-buildkite-analytics could not establish an initial connection with Buildkite due to #{e.message} after #{MAX_RECONNECTION_ATTEMPTS} attempts. You may be missing some data for this test suite, please contact support if this issue persists."
         else
           sleep(WAIT_BETWEEN_RECONNECTIONS)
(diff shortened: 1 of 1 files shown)


jasmine-q (approved):
Thanks for the easy to understand PR description, code seems straightforward too (a noob when it comes to threads)

lib/rspec/buildkite/analytics/session.rb
```diff
@@ -52,9 +52,22 @@ def initialize(url, authorization_header, channel)
...
-      connect
-    rescue TimeoutError, InitialConnectionFailure => e
-      $stderr.puts "rspec-buildkite-analytics could not establish an initial connection with Buildkite due to #{e.message}. You may be missing some data for this test suite, please contact support."
+      reconnection_count = 0
+
+      begin
+        reconnection_count += 1
+        connect
+      rescue TimeoutError, InitialConnectionFailure => e
+        @logger.write("failed initial connection attempt #{reconnection_count} due to #{e}")
```
blaknite, line 61: I think if this code gets restructured a bit we can provide some better debug messages like the one here and in my other comments.

```suggestion
        @logger.write("rspec-buildkite-analytics could not establish an initial connection with Buildkite due to #{e}. Attempting retry #{reconnection_count} of #{MAX_RECONNECTION_ATTEMPTS}...")
```

lib/rspec/buildkite/analytics/session.rb
```diff
@@ -52,9 +52,22 @@ def initialize(url, authorization_header, channel)
...
+        if reconnection_count > MAX_RECONNECTION_ATTEMPTS
+          $stderr.puts "rspec-buildkite-analytics could not establish an initial connection with Buildkite due to #{e.message}. You may be missing some data for this test suite, please contact support if this issue persists."
```
blaknite, line 63:
```suggestion
          $stderr.puts "rspec-buildkite-analytics could not establish an initial connection with Buildkite due to #{e.message} after #{MAX_RECONNECTION_ATTEMPTS} attempts. You may be missing some data for this test suite, please contact support if this issue persists."
```

blaknite (commented):
A great improvement 🙌

Wondering if we can make the debug messages a bit better. Otherwise no issues with how it works.

PR #77, opened 2022-01-18
Store collector version

niceking:
We talked vaguely in the planning meeting about storing the collector version in the upload jsons S3 file but I thought that 
maybe it was easier/more efficient to just put it in the run env? Store it once per run rather than once upload part? It's also less code changes this way 😂

### DIFF
--- a/lib/rspec/buildkite/analytics/ci.rb
+++ b/lib/rspec/buildkite/analytics/ci.rb
@@ -15,14 +15,16 @@ module RSpec::Buildkite::Analytics::CI
         "job_id" => ENV["BUILDKITE_JOB_ID"],
         "message" => ENV["BUILDKITE_MESSAGE"],
         "debug" => ENV["BUILDKITE_ANALYTICS_DEBUG_ENABLED"],
-        "version" => RSpec::Buildkite::Analytics::VERSION
+        "version" => RSpec::Buildkite::Analytics::VERSION,
+        "collector" => RSpec::Buildkite::Analytics::NAME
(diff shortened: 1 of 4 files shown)


lib/rspec/buildkite/analytics/ci.rb
```diff
@@ -14,13 +14,15 @@ def self.env
         "number" => ENV["BUILDKITE_BUILD_NUMBER"],
         "job_id" => ENV["BUILDKITE_JOB_ID"],
         "message" => ENV["BUILDKITE_MESSAGE"],
-        "debug" => ENV["BUILDKITE_ANALYTICS_DEBUG_ENABLED"]
+        "debug" => ENV["BUILDKITE_ANALYTICS_DEBUG_ENABLED"],
+        "version" => RSpec::Buildkite::Analytics::VERSION
```
blaknite, line 18: As new collectors show up we could get conflicting versions. I'm wondering if this should be
```suggestion
        "collector" => "RSpec - #{RSpec::Buildkite::Analytics::VERSION}"
```
or something?

blaknite: or maybe a new key of "collector" that has "RSpec in it.

niceking:
Cool I've opted it add in a new "collector" field. Made it "rspec-buildkite" incase someone decides they don't like our one in the future 🤷‍♀️ 

Will do a gem release after this gets merged

jasmine-q (approved):

PR #78, opened 2022-02-07
Add support for annotation spans

blaknite:
This PR adds support for basic annotations. Annotations a special kind of span that will be supported by Test Analytics to facilitate easier debugging of threaded code within a test execution.

An annotation can be created by calling `RSpec::Buildkite::Analytics.annotate("some kind of annotation string")`

It'll then show as something like this in the UI.

Screen Shot 2022-02-07 at 11 08 02 am
.

### DIFF
--- a/lib/rspec/buildkite/analytics.rb
+++ b/lib/rspec/buildkite/analytics.rb
@@ -30,4 +30,10 @@ module RSpec::Buildkite::Analytics
 
     self::Uploader.configure
   end
+
+  def self.annotate(content)
+    tracer = RSpec::Buildkite::Analytics::Uploader.tracer
+    tracer&.enter("annotation", **{ content: content })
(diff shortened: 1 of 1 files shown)


jasmine-q:
Should we be setting a max annotation length or anything? 🤔

blaknite:
> Should we be setting a max annotation length or anything? 

Currently we don't have any max lengths for any of the span payloads. We have a ticket to address database column sizes, but these would only ever be stored in S3 so it's a 🤷‍♂️ from me. I'll pop a message in basecamp. I don't think it's required to ship this PR.

jasmine-q (approved):

niceking (approved):
LGTM, agreed that any discussion about limits on this data is not a blocker for this!

PR #80, opened 2022-02-16
Support multiple CI platforms and generic env

blaknite:
This PR reinstates the support for GitHub Actions and CircleCI. It also adds support for override environment variables in the `BUILDKITE_ANALYTICS_` namespace.

If `GITHUB_RUN_NUMBER` is set then we will apply some sensible defaults for the run env based on those provided by GitHub Actions. If `CIRCLE_BUILD_NUM` is set we do the same for CircleCI. When the generic `CI` environment variable is set the run env "ci" is set to "generic".

When a URL and NUMBER are provided a link to the build will show in the Test Analytics UI regardless of CI platform. This also applies when the `CI` env is not set but `BUILDKITE_ANALYTICS_URL` and `BUILDKITE_ANALYTICS_NUMBER` are set.

Screen Shot 2022-02-16 at 8 25 57 pm

`BUILDKITE_ANANLYTICS_MESSAGE` can be used within CI platforms that do not have an environment variable that is equivalent to `BUILDKITE_MESSAGE`.

Supported variables:
`BUILDKITE_ANALYTICS_KEY` - a unique key for the build that initiated the Test Analytics run.
`BUILDKITE_ANALYTICS_URL` - a URL that links to the build.
`BUILDKITE_ANALYTICS_BRANCH` - the branch or ref that this run is for.
`BUILDKITE_ANALYTICS_SHA` - the commit sha for the head of the branch.
`BUILDKITE_ANALYTICS_NUMBER` - the build number of the build.
`BUILDKITE_ANALYTICS_JOB_ID` - the id of a job within the build.
`BUILDKITE_ANANLYTICS_MESSAGE` - the commit message for the head of the branch.

### DIFF
--- a/lib/rspec/buildkite/analytics/ci.rb
+++ b/lib/rspec/buildkite/analytics/ci.rb
@@ -2,30 +2,85 @@
 
 require "securerandom"
 
-module RSpec::Buildkite::Analytics::CI
+class RSpec::Buildkite::Analytics::CI
   def self.env
-    if ENV["BUILDKITE_BUILD_ID"]
-      {
-        "CI" => "buildkite",
-        "key" => ENV["BUILDKITE_BUILD_ID"],
-        "url" => ENV["BUILDKITE_BUILD_URL"],
-        "branch" => ENV["BUILDKITE_BRANCH"],
-        "commit_sha" => ENV["BUILDKITE_COMMIT"],
-        "number" => ENV["BUILDKITE_BUILD_NUMBER"],
-        "job_id" => ENV["BUILDKITE_JOB_ID"],
-        "message" => ENV["BUILDKITE_MESSAGE"],
-        "debug" => ENV["BUILDKITE_ANALYTICS_DEBUG_ENABLED"],
-        "version" => RSpec::Buildkite::Analytics::VERSION,
-        "collector" => RSpec::Buildkite::Analytics::NAME
-      }
-    else
-      {
-        "CI" => nil,
-        "key" => SecureRandom.uuid,
-        "debug" => ENV["BUILDKITE_ANALYTICS_DEBUG_ENABLED"],
-        "version" => RSpec::Buildkite::Analytics::VERSION,
-        "collector" => RSpec::Buildkite::Analytics::NAME
-      }
-    end
+    new.env
+  end
+
+  # The analytics env are more specific than the automatic ci platform env.
+  # If they've been specified we'll assume the user wants to use that value instead.
+  def env
+    ci_env.merge(analytics_env)
+  end
+
+  private
+
+  def ci_env
+    return buildkite if ENV["BUILDKITE_BUILD_ID"]
+    return github_actions if ENV["GITHUB_RUN_NUMBER"]
+    return circleci if ENV["CIRCLE_BUILD_NUM"]
+    return generic if ENV["CI"]
+
+    {
+      "CI" => nil,
+      "key" => SecureRandom.uuid,
+    }
+  end
+
+  def analytics_env
+    {
+      "key" => ENV["BUILDKITE_ANALYTICS_KEY"],
+      "url" => ENV["BUILDKITE_ANALYTICS_URL"],
+      "branch" => ENV["BUILDKITE_ANALYTICS_BRANCH"],
+      "commit_sha" => ENV["BUILDKITE_ANALYTICS_SHA"],
+      "number" => ENV["BUILDKITE_ANALYTICS_NUMBER"],
+      "job_id" => ENV["BUILDKITE_ANALYTICS_JOB_ID"],
+      "message" => ENV["BUILDKITE_ANANLYTICS_MESSAGE"],
+      "debug" => ENV["BUILDKITE_ANALYTICS_DEBUG_ENABLED"],
+      "version" => RSpec::Buildkite::Analytics::VERSION,
+      "collector" => RSpec::Buildkite::Analytics::NAME,
+  }.compact
+  end
+
+  def generic
+    {
+      "CI" => "generic",
+      "key" => SecureRandom.uuid,
+    }
+  end
+
+  def buildkite
+    {
+      "CI" => "buildkite",
+      "key" => ENV["BUILDKITE_BUILD_ID"],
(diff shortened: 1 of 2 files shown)


lib/rspec/buildkite/analytics/ci.rb
```diff
@@ -2,30 +2,83 @@
...
+      "url" => ENV["BUILDKITE_ANALYTICS_URL"],
+      "branch" => ENV["BUILDKITE_ANALYTICS_BRANCH"],
+      "commit_sha" => ENV["BUILDKITE_ANALYTICS_SHA"],
+      "number" => ENV["BUILDKITE_ANALYTICS_NUMBER"],
+      "job_id" => ENV["BUILDKITE_ANALYTICS_JOB_ID"],
+      "message" => ENV["BUILDKITE_ANANLYTICS_MESSAGE"],
+      "debug" => ENV["BUILDKITE_ANALYTICS_DEBUG_ENABLED"],
+      "version" => RSpec::Buildkite::Analytics::VERSION,
+      "collector" => RSpec::Buildkite::Analytics::NAME,
+    }.reject { |k, v| v.nil? }
```
niceking, line 40: `.compact` does the same thing? Is this an ActiveSupport thing again 🤔

niceking: No I'm pretty sure I've used `.compact` in other places in this code base...

lib/rspec/buildkite/analytics/ci.rb
```diff
@@ -2,30 +2,83 @@
...
-        "debug" => ENV["BUILDKITE_ANALYTICS_DEBUG_ENABLED"],
-        "version" => RSpec::Buildkite::Analytics::VERSION,
-        "collector" => RSpec::Buildkite::Analytics::NAME
-      }
-    end
+    new.env
+  end
+
+  def env
+    ci_env.merge(default)
```
niceking, line 11: I found this a little hard to grok as my initial reaction was "whoa we'll be overwriting the ci specific env with the default!!" and then I read on and you're removing the nil keys so that doesn't happen. But perhaps `default.merge(ci_env)` might be clearer from the get, and you won't have to remove nil keys?

lib/rspec/buildkite/analytics/ci.rb
blaknite: Yep you can. TIL you can compact a hash. Nice!

lib/rspec/buildkite/analytics/ci.rb
```diff
@@ -2,30 +2,83 @@
...
+  private
+
+  def ci_env
+    return buildkite if ENV["BUILDKITE_BUILD_ID"]
+    return github_actions if ENV["GITHUB_RUN_NUMBER"]
+    return circleci if ENV["CIRCLE_BUILD_NUM"]
+    return generic if ENV["CI"]
+
+    {
+      "CI" => nil,
```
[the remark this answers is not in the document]
blaknite: I was thinking that it might be work us knowing if something was run in CI vs dev. Am I overthinking it?

lib/rspec/buildkite/analytics/ci.rb
blaknite: I'll rename it. They're not really default, they're more specific. That is, if someone explicitly specifies a `BUILDKITE_ANALYTICS_*` environment variable they're likely wanting to use that value.

lib/rspec/buildkite/analytics/ci.rb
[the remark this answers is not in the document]
niceking: Ah yeah that seems like a fair enough use case!

lib/rspec/buildkite/analytics/ci.rb
```diff
@@ -2,30 +2,85 @@
...
+      "message" => ENV["BUILDKITE_ANANLYTICS_MESSAGE"],
+      "debug" => ENV["BUILDKITE_ANALYTICS_DEBUG_ENABLED"],
+      "version" => RSpec::Buildkite::Analytics::VERSION,
+      "collector" => RSpec::Buildkite::Analytics::NAME,
+  }.compact
+  end
+
+  def generic
+    {
+      "CI" => "generic",
```
niceking, line 47:
```suggestion
      "CI" => ENV["CI"],
```
Generic providers still provide a value for this key, yes? Should we show whatever their CI name is here?

lib/rspec/buildkite/analytics/ci.rb
```diff
@@ -2,30 +2,85 @@
...
-        "debug" => ENV["BUILDKITE_ANALYTICS_DEBUG_ENABLED"],
-        "version" => RSpec::Buildkite::Analytics::VERSION,
-        "collector" => RSpec::Buildkite::Analytics::NAME
-      }
-    end
+    new.env
+  end
+
+  # The analytics env are more specific than the automatic ci platform env.
+  # If they've been specified we'll assume the user wants to use that value instead.
```
niceking, line 11: Great comment!!!

lib/rspec/buildkite/analytics/ci.rb
blaknite: From what I've seen, that value is often "true". Not really the best descriptor for a CI type.

niceking: What do you mean by "from what I have seen"? Like we don't currently have any customers who just use Test Analytics without Pipelines right? I was assuming that we were going to be launching support for other CI provider with this PR, and that there'd be some instructions like "Set the CI env variable to the name of your CI tool" as that's what all the other values of CI are?

blaknite: Buildkite, GHA and CircleCI all set this to "true". In absence of further research I'm assuming others do as well.

niceking: Okkkk right so what you mean is that in their docs GHA and CircleCI both specific that they set CI to true and that this is probably a general pattern that other CI providers follow - was not clear that that's where you were looking!

niceking (approved):

PR #81, opened 2022-03-23
Add hooks to allow the use of different test frameworks

mariovisic:
Currently the library is fairly strongly tied to RSpec. In order to support
different ruby testing frameworks it makes sense to re-use the logic in this
library around reporting and sockets, rather than creating separate gems for
each testing framework.

Adding a hook system will allow different code to setup each test framework
that we want the gem to work with, I'll be working on minitest support after
these changes are merged.

The structure of the hooks is lifted from the VCR gem which already uses
this mechanism. In this change I've updated the library to only require
RSpec as a development dependency as with each added hook the
dependencies for the gem will grow and there's no use in having projects
rely on multiple test frameworks.

If we want to go down this path it would make sense to rename the library to
`buildkite-analytics` (which seems available on rubygems) as it'll no longer be
RSpec specific and create a plan to migrate existing users of the
`rspec-buildkite-analytics` library over to the new one.


blaknite:
Thanks @mariovisic!!

> Adding a hook system will allow different code to setup each test framework
that we want the gem to work with, I'll be working on minitest support after
these changes are merged.

Awesome, this is the direction I was thinking of heading as well. Awesome to see it coming together 🙌

> I've updated the library to only require
RSpec as a development dependency as with each added hook the
dependencies for the gem will grow and there's no use in having projects
rely on multiple test frameworks.

Nice!

> If we want to go down this path it would make sense to rename the library to
buildkite-analytics (which seems available on rubygems) as it'll no longer be
RSpec specific and create a plan to migrate existing users of the
rspec-buildkite-analytics library over to the new one.

There was some talk about merging this with the `buildkite` gem as well. I'll raise it with the team and get back to you with our decision. Either way, keen to follow this approach.

We're not currently building PRs from forks so I'll need to check the config and update it so we can get a build for this PR. In the meantime I'll pull this down and give it a test locally.

Thanks again!

blaknite (approved):

PR #85, opened 2022-04-03
Replace invalid UTF-8 characters in test names

mariovisic:
In the `Session` class we convert our test results into a JSON string to
send over the wire. Calling `to_json` on a string that has invalid UTF-8
characters result in a `JSON::GeneratorError`. Which means that if a
test name, or any other data we generate has any invalid characters we
won't be able to send the results to analytics. So we replace any of
these characters.

I think it would have been good to avoid the recursive function here and
only re-encoded the strings that are generated from the users test
suite, but that happens in quite a few places, it would be easy to skip
doing this when adding additional data from the test suite as well; also
those values appearing in deep nested structures, so we can't just call
that function for a few keys, we need to recursively visit each string
in our hash.

### DIFF
--- a/lib/rspec/buildkite/analytics/uploader.rb
+++ b/lib/rspec/buildkite/analytics/uploader.rb
@@ -42,7 +42,7 @@ module RSpec::Buildkite::Analytics
       end
 
       def as_hash
-        {
+        strip_invalid_utf8_chars(
           id: @id,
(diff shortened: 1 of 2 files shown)


keithpitt (approved):

blaknite (approved):
Thanks @mariovisic! I'll get this one merged in 🙌

PR #89, opened 2022-04-21
Add security advice to API key config in readme

joshprice:
The original instructions seem to imply that you should add your API key directly to a config line in your repo. Which is generally considered unwise. Probably best to guide users well away from this option if at all possible.

This is just a draft proposal so please change/expand/discuss!

/cc @toolmantim @keithpitt

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,14 @@ group :test do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


joshprice:
Looks like this behaviour might need to be explained in the docs too?

https://github.com/buildkite/rspec-buildkite-analytics/blob/main/lib/rspec/buildkite/analytics.rb#L24

joshprice:
Just found that this is already described well over here https://buildkite.com/docs/test-analytics/integrations#integrating-with-rspec so will update this PR with instructions that mirror those.

joshprice:
Closed in favour of #90

PR #90, opened 2022-04-23
Update README installation instructions

joshprice:
Updated installation instruction to match https://buildkite.com/docs/test-analytics/integrations#integrating-with-rspec

This also fixes the security raised in #89 and replaces that PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,32 +4,46 @@ This gem collects data about your test suite's performance and reliability, and 
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


gchan (approved):
Thanks for your contribution! 🌟

JuanitoFatas (approved):
Thanks! I have changed the example env name to prefix with `BUILDKITE` so people know where it is coming from :)

PR #93, opened 2022-05-05
Retransmit unconfirmed idents

blaknite:
During session close we send an EOT message that triggers a confirm response containing all idents accepted by the server. If, after confirmation, there are some idents remain unconfirmed we now retransmit them, send another EOT and await their confirmation.

This fixes an issue where we would wait the full 75 second timeout if any idents remained unconfirmed.

To further improve issues with long timeouts, I have added three new environment variables:

- `BUILDKITE_ANALYTICS_CONFIRMATION_TIMEOUT` - seconds to wait for ident confirmation during session close
- `BUILDKITE_ANALYTICS_RECONNECTION_ATTEMPTS` - number of reconnection attempts if session is unexpectedly closed
- `BUILDKITE_ANALYTICS_RECONNECTION_WAIT` - seconds to wait between reconnection attempts

I've also added some status messages for connect, waiting for confirm, and completed. This should give users a better understanding of if we're holding up their test runs.

### DIFF
--- a/lib/rspec/buildkite/analytics/session.rb
+++ b/lib/rspec/buildkite/analytics/session.rb
@@ -116,14 +116,14 @@ module RSpec::Buildkite::Analytics
       # to which the server will respond with the last bits of data
       send_eot
 
+      # After EOT, we wait for 75 seconds for the send queue to be drained and for the
+      # server to confirm the last idents. If everything has already been confirmed we can
+      # proceed without waiting.
       @idents_mutex.synchronize do
-        @logger.write("waiting for last confirm")
-        # Here, we sleep for 75 seconds while waiting for the send
-        # queue to be drained and for the server to confirm the last
-        # idents.
-        # We are woken up when the unconfirmed_idents is empty, and given back the mutex to
-        # continue operation.
-        @empty.wait(@idents_mutex, CONFIRMATION_TIMEOUT) unless @unconfirmed_idents.empty?
+        if @unconfirmed_idents.any?
+          @logger.write("waiting for last confirm")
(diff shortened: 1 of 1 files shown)


lib/rspec/buildkite/analytics/session.rb
```diff
@@ -255,23 +260,21 @@ def queue_and_track_result(ident, result_as_hash)
...
         }
-
-        @send_queue << data
       end
     end
 
-    def remove_unconfirmed_idents(idents)
-      return if idents.empty?
+    def confirm_idents(idents)
+      retransmit_required = @closing
```
JuanitoFatas, line 271: Should we still keep the early return when `idents` is empty?

lib/rspec/buildkite/analytics/session.rb
```diff
@@ -255,23 +260,21 @@ def queue_and_track_result(ident, result_as_hash)
...
 
       @idents_mutex.synchronize do
         # Remove received idents from unconfirmed_idents
         idents.each { |key| @unconfirmed_idents.delete(key) }
 
-        @logger.write("received confirm for indentifiers: #{idents.join(", ")}")
+        @logger.write("received confirm for indentifiers: #{idents}")
```
JuanitoFatas, line 277: Why remove the `join(", ")`? I guess separate UUIDs by comma is not much gain from looking at the array itself.

JuanitoFatas (approved):
Try this branch locally and it works 👍🏼

gchan (approved):
🚀 LGTM!

PR #96, opened 2022-05-16
Add env var to configure upload messages

gough:
Hi there!

We just started using this RSpec plugin, but the waiting/connected messages are getting in the way of reading the test output. This PR adds an additional environment variable `BUILDKITE_ANALYTICS_UPLOAD_MESSAGES` to optionally disable them (defaults to enabled).

Please let me know what you think! Happy to make any requested changes to get this merged quickly.


blaknite:
Thanks @gough! Keen as to make our output a bit quieter.

@JuanitoFatas has opened #97 to introduce a more robust and configurable logging setup. Once that's merged you should be able to configure the gem to the desired log level and squash these messages 😊

PR #98, opened 2022-05-20
Make trace lookup faster by using a hash

blaknite:
Backport from #86.

`RSpec::Buildkite::Analytics.uploader.traces` was an array and we would loop through this many times to find our matching trace. On larger test suites with 1000s of tests, it would cause 100s of 1000s of loops. We can instead lookup the value directly with a hash.


JuanitoFatas (approved):
BIG `O(1)` lookup by Hash is 😍 

bigo

mariovisic:
thanks @blaknite for getting this in 👍

PR #100, opened 2022-05-23
Fix namespace of Reporter

JuanitoFatas:
The [reporter class](https://github.com/buildkite/rspec-buildkite-analytics/blob/f5636f04ba1c0ed1add2ef43d51a2216e5d15af7/lib/buildkite/collector/rspec_plugin/reporter.rb#L4) is under `RSpecPlugin`. Spotted when test failed:

CleanShot 2022-05-23 at 10 54 29@2x


blaknite:
This is what I get for running this against a test suite with 100% passing tests.

How hard is it to improve test coverage for this?

blaknite (approved):
✨

PR #101, opened 2022-05-24
Improve user experience running the collector

JuanitoFatas:
We hit an `EOFError` that happens frequently, and before we could fix that, use `WARN` instead of `ERROR` to improve user experience using our collector.

The specific `EOFError` was:

```
EOFError: end of file reached
lib/ruby/3.0.0/openssl/buffering.rb:148:in `sysread'
lib/ruby/3.0.0/openssl/buffering.rb:148:in `readpartial'
rspec-buildkite-analytics/lib/buildkite/collector/socket_connection.rb:73:in `block in initialize'
```

- [x] I’ve verified this works by manual testing

### DIFF
--- a/lib/buildkite/collector/socket_connection.rb
+++ b/lib/buildkite/collector/socket_connection.rb
@@ -76,7 +76,14 @@ module Buildkite::Collector
             @session.handle(self, data.data)
           end
         end
-      rescue EOFError, Errno::ECONNRESET => e
+      rescue EOFError => e
+        Buildkite::Collector.logger.warn("#{e}")
+        if @socket
+          Buildkite::Collector.logger.warn("attempting disconnected flow")
+          @session.disconnected(self)
+          disconnect
+        end
+      rescue Errno::ECONNRESET => e
         Buildkite::Collector.logger.error("#{e}")
         if @socket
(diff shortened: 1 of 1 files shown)


gchan (approved):
LGTM! 

Do we want to be more specific and only `warn` for that particular subset of `EOFError` (specifically 'end of file reached'). I'm uncertain about the feasibility/ease of doing this however. I'm also unsure of other `EOFError` errors that might occur where we may want to use `error` rather than `warn`.

JuanitoFatas:
> Do we want to be more specific and only `warn` for that particular subset of `EOFError` (specifically 'end of file reached'). I'm uncertain about the feasibility/ease of doing this however. I'm also unsure of other `EOFError` errors that might occur where we may want to use `error` rather than `warn`.

I think logging with `WARN` already provides has the ability to understand what’s going on ([puma rescued no log](https://github.com/puma/puma/blob/3d24dc5936d0e570376fa843a682cf2c5481041e/lib/puma/client.rb#L217-L218)) if we turned on [debug flags](https://github.com/buildkite/rspec-buildkite-analytics/wiki/Logging#turning-on-debug-mode), so I would not do specifc the error handling here.

PR #102, opened 2022-05-25
Handle re-raised disconnected exceptions

JuanitoFatas:
When the `disconnected` attempt exhausted, we re-raise the exception that bubbles up back to the socket connection. People may have installed the Ruby collector on an Android project, and in this case, we should fail gracefully.

### DIFF
--- a/lib/buildkite/collector/socket_connection.rb
+++ b/lib/buildkite/collector/socket_connection.rb
@@ -76,6 +76,9 @@ module Buildkite::Collector
             @session.handle(self, data.data)
           end
         end
+      # These get re-raise from session, we should fail gracefully
+      rescue *Buildkite::Collector::Session::DISCONNECTED_EXCEPTIONS => e
+        Buildkite::Collector.logger.error("We could not establish a connection with Buildkite Test Analytics. The error was: #{e.message}. If this is a problem, please contact support.")
       rescue EOFError => e
         Buildkite::Collector.logger.warn("#{e}")
         if @socket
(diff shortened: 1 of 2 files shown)


lib/buildkite/collector/socket_connection.rb
```diff
@@ -76,6 +76,9 @@ def initialize(session, url, headers)
             @session.handle(self, data.data)
           end
         end
+      # These get re-raise from session, we should fail gracefully
+      rescue Buildkite::Collector::Session::DISCONNECTED_EXCEPTIONS => e
```
blaknite, line 80: I agree this is the right thing to do, however I'm not sure if we need to attempt to execute the disconnect flow. Thoughts?

JuanitoFatas: > if we need to attempt to execute the disconnect flow.

??? Here we rescue the exception raised from `Session`, only logs an error, did not execution the disconnect flow.

This is attempt execute disconnect flow looks like in the `rescue IndexError` block:

https://github.com/buildkite/rspec-buildkite-analytics/blob/fbce8caa2b535a4cf1098222026abb045d569e66/lib/buildkite/collector/socket_connection.rb#L134-L138

Am I missing something 🤔 😅

swebb: @JuanitoFatas I couldn't really follow what you're saying.

@blaknite if I'm understanding it correctly I don't think we want to execute the disconnect flow. The exceptions it's catching are "DISCONNECTED_EXCEPTIONS", i.e., we have already tried to perform the disconnect flow and it raised an error. Now we should print an error and gracefully handle the exception.

swebb:
I don't feel confident enough with this code to approve it, but it looks good to me. 👍

blaknite (approved):

PR #105, opened 2022-06-01
Extract into a HTTP Client

JuanitoFatas:
This merges into #103.

This will make changes to HTTP easier in the future, also easier to mock in tests.


swebb (approved):
👍

PR #108, opened 2022-06-02
Detailed tracing can be disabled

blaknite:
Sometimes people might run into issues where the detailed tracing is incompatible with their tests or is causing tests to run slower. To make it so they can keep using Test Analytics I've added the option to disable the detailed tracing.

```ruby
Buildkite::TestCollector.configure(
  ...
  tracing_enabled: false
)
```


swebb (approved):
🚢

Issue #111, opened 2022-06-02
Minitest with Knapsack


paulca:
Hi,

I just installed the test collector in our app, following the instructions on the site.

I ran in to an initial issue that `WebMock` wasn't allowing requests, and then I needed to move the `WebMock` `allow` config above `Buildkite::TestCollector.configure(hook: :minitest)`.

This got the tests running, but now I'm getting an error in the internals:

```

# Running:
--
  |  
  | Capybara starting Puma...
  | * Version 5.6.4 , codename: Birdie's Version
  | * Min threads: 0, max threads: 4
  | * Listening on http://172.16.112.5:3000
  | rake aborted!
  | TypeError: no implicit conversion of Pathname into String
  | /bundle/ruby/2.7.0/gems/buildkite-test_collector-1.0.0/lib/buildkite/test_collector/minitest_plugin/trace.rb:60:in `delete_prefix'
  | /bundle/ruby/2.7.0/gems/buildkite-test_collector-1.0.0/lib/buildkite/test_collector/minitest_plugin/trace.rb:60:in `file_name'
  | /bundle/ruby/2.7.0/gems/buildkite-test_collector-1.0.0/lib/buildkite/test_collector/minitest_plugin/trace.rb:53:in `location'
  | /bundle/ruby/2.7.0/gems/buildkite-test_collector-1.0.0/lib/buildkite/test_collector/minitest_plugin/trace.rb:40:in `as_hash'
  | /bundle/ruby/2.7.0/gems/buildkite-test_collector-1.0.0/lib/buildkite/test_collector/session.rb:145:in `write_result'
  | /bundle/ruby/2.7.0/gems/buildkite-test_collector-1.0.0/lib/buildkite/test_collector/minitest_plugin/reporter.rb:15:in `record'
  | /bundle/ruby/2.7.0/gems/minitest-5.15.0/lib/minitest.rb:874:in `block in record'
  | /bundle/ruby/2.7.0/gems/minitest-5.15.0/lib/minitest.rb:873:in `each'
  | /bundle/ruby/2.7.0/gems/minitest-5.15.0/lib/minitest.rb:873:in `record'
  | /bundle/ruby/2.7.0/gems/minitest-5.15.0/lib/minitest.rb:350:in `run_one_method'
  | /bundle/ruby/2.7.0/gems/minitest-5.15.0/lib/minitest.rb:337:in `block (2 levels) in run'
[...]

PR #112, opened 2022-06-03
Fix `project_dir` issue expecting a string, not a `Pathname`

paulca:
**Issue**
See #111, this lib seems to expect a string for the `project_dir`, but `Rails.root` returns `Pathname`

**Steps to reproduce**
Run minitest with the gem enabled in a Rails 6+ app

**Expected**
The gem should do its job and report analytics back to Buildkite

**Actual**
See #111 for the output

As suggested by @nickvanw, this PR simply calls `to_s` on `Rails.root` in the `project_dir` method in `minitest_plugin/trace.rb`

I added a test too that fails without the fix.

### DIFF
--- a/lib/buildkite/test_collector/minitest_plugin/trace.rb
+++ b/lib/buildkite/test_collector/minitest_plugin/trace.rb
@@ -66,7 +66,7 @@ module Buildkite::TestCollector::MinitestPlugin
 
     def project_dir
       if defined?(Rails) && Rails.respond_to?(:root)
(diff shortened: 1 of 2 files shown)


njcameron:
Just bumped into this issue locally too

JuanitoFatas (approved):
Looks good to me!

JuanitoFatas:
Thank you so much for fixing this and adding tests, this is great!

PR #97, opened 2022-05-19
Introduce Logger utility

JuanitoFatas:
## Summary

This Pull Request adds a `Logger` class to replace `puts`, `$stderr.puts` and custom `logger.write` usages, which replace #96 that if we want to suppress unwanted log messages, set the logger to send logs to `/dev/null`:

```ruby
RSpec::Buildkite::Analytics.logger = Logger.new("/dev/null")
```

#### Detailed changes

- Introduce a Logger class which can be changed by user
- The logs can be enabled on by env `BUILDKITE_ANALYTICS_DEBUG_ENABLED` or standard Ruby flag `$DEBUG`
- default log level is `::Logger::WARN`
- The log has ISO 8601 timestamp with 9 digit precision, Process id and current Thread id
  `2022-05-19T07:43:52.795221000Z pid=1977 tid=7840 DEBUG: starting socket connection process`
- Replace all `puts`, `$stderr.puts` with `RSpec::Buildkite::Analytics.logger.debug` — Should this be `INFO` instead 🤔? To keep it simple, I replace with `.debug`.
- Replace all `logger.write` with `RSpec::Buildkite::Analytics.logger.debug`
- Do we still want to keep the `Reporter` class dump logs in a session to a file feature? If so, we will keep [this class](https://github.com/buildkite/rspec-buildkite-analytics/blob/69f7bd56cc07133387bc0b9fc4cb8b3b3cce020a/lib/rspec/buildkite/analytics/session.rb#L15-L33), but only for dumping session logs into a file. 

- [x] I’ve verified this Pull Request works with production

  <details>
  <summary>View logs</summary>

  ```
  Randomized with seed 44769
  2022-05-19T07:43:52.795221000Z pid=1977 tid=7840 DEBUG: starting socket connection process
  2022-05-19T07:43:53.541832000Z pid=1977 tid=22100 DEBUG: listening in on socket
  2022-05-19T07:43:53.561479000Z pid=1977 tid=22100 DEBUG: received welcome
  2022-05-19T07:43:53.758843000Z pid=1977 tid=22100 DEBUG: received confirm_subscription
  2022-05-19T07:43:53.760751000Z pid=1977 tid=7840 DEBUG: Connected to Buildkite Test Analytics!
  2022-05-19T07:43:53.760814000Z pid=1977 tid=7840 DEBUG: connected
  2022-05-19T07:43:53.764155000Z pid=1977 tid=22120 DEBUG: hello from write thread
[...]

PR #115, opened 2022-06-08
Remove unused debug_filepath

JuanitoFatas:
Thanks to @pda pointed out in https://github.com/buildkite/test-collector-ruby/pull/97#issuecomment-1147098900.

Remove unused `debug_filepath`.

This was introduced at https://github.com/buildkite/test-collector-ruby/commit/8d99faf74dbf505476202c2ab67e5a0cdd493c76 to dump debug logs into a preferred location but it is not in used anymore

I do not believe anyone but us uses this feature but this is a breaking API change for the `configure` method.

### DIFF
--- a/lib/buildkite/test_collector.rb
+++ b/lib/buildkite/test_collector.rb
@@ -19,15 +19,13 @@ module Buildkite
       attr_accessor :uploader
       attr_accessor :session
       attr_accessor :debug_enabled
-      attr_accessor :debug_filepath
       attr_accessor :tracing_enabled
     end
 
-    def self.configure(hook:, token: nil, url: nil, debug_enabled: false, debug_filepath: nil, tracing_enabled: true)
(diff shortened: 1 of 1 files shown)


gchan (approved):
> I do not believe anyone but us uses this feature but this is a breaking API change for the configure method.

Is it breaking enough for warrant a minor version bump? 

It is unlikely anyone would be using this debugging feature however.

JuanitoFatas:
> Is it breaking enough to warrant a minor version bump?

If we follow SemVer, I think this should be major version bump 😬 

> It is unlikely anyone would be using this debugging feature however.

Yes...

pda:
> > Is it breaking enough to warrant a minor version bump?
> 
> If we follow SemVer, I think this should be major version bump 😬

Heh, in that case you could keep the `debug_filepath: nil` in the method args (assuming that method is part of a public API?) and emit a deprecation warning if it receives a value until the next major version.

PR #121, opened 2022-06-10
Implement Test Location Prefix 

JuanitoFatas:
This Pull Request ports JavaScript Collector feature https://github.com/buildkite/test-collector-javascript/pull/30.

This allows a Monolith repository to configure different test name prefix via environment variable `BUILDKITE_ANALYTICS_LOCATION_PREFIX` across multiple pipelines.

### DIFF
--- a/lib/buildkite/test_collector/minitest_plugin/trace.rb
+++ b/lib/buildkite/test_collector/minitest_plugin/trace.rb
@@ -58,7 +58,14 @@ module Buildkite::TestCollector::MinitestPlugin
     alias_method :identifier, :location
 
     def file_name
-      @file_name ||= File.join('./', source_location[0].delete_prefix(project_dir))
+      @file_name ||= begin
+        prefix = ENV["BUILDKITE_ANALYTICS_LOCATION_PREFIX"]
+        if prefix && !prefix.empty?
+          File.join(prefix, source_location[0].delete_prefix(project_dir))
+        else
+          File.join('./', source_location[0].delete_prefix(project_dir))
+        end
+      end
     end
 
     def line_number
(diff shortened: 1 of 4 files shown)


toolmantim:
Nice! The JS one also reports `run_env[location_prefix]` in the data sent too.

lib/buildkite/test_collector/minitest_plugin/trace.rb
```diff
@@ -58,7 +58,14 @@ def location
     alias_method :identifier, :location
 
     def file_name
-      @file_name ||= File.join('./', source_location[0].delete_prefix(project_dir))
+      @file_name ||= begin
```
gchan, line 61: Should we only be prefixing the `location` and leaving the `file_name` unchanged?

I noticed in the RSpec and JS versions, `file_name` is unchanged and only `location` is modified?

JuanitoFatas: Yea I started with changing location. Say someone added a prefix `payments`, changing `location` will result as:

`"payments/./path/to/the/test"`

because of here https://github.com/buildkite/test-collector-ruby/blob/4a1d0fb0b0f365758bf3c1b77500aa22dcbe2743/lib/buildkite/test_collector/minitest_plugin/trace.rb#L61

While changing `file_name` will avoid that and returns:

`"payments/path/to/the/test"`

and I think it makes more sense to change `file_name` here.

What do you think?

gchan: Thanks for pointing this out! The `./` doesn't make sense when prefixing!

I just re-read the change in the JS version and it looks like both file_name and location are prefixed in that version. 

https://github.com/buildkite/test-collector-javascript/blob/6abcf44df181d8fe377be481f0fdc894306f547d/jest/reporter.js#L84-L85=

For consistency, should we prefix both file and location everywhere?

JuanitoFatas: > should we prefix both file and location everywhere?

👍🏼  will do, thanks!

JuanitoFatas:
will revisit this later

PR #125, opened 2022-06-16
Loading minitest without initializing the plugin should work

paulca:
**Issue**
Requiring the gem but not running `Buildkite::TestCollector.configure(hook: :minitest)` causes an exception:

CleanShot 2022-06-16 at 15 41 59@2x

**Steps to reproduce**
- In a Rails 7 app
- Add `buildkite-test_collector` to the `Gemfile`
- Do not add ```Buildkite::TestCollector.configure(hook: :minitest)```

**Expected**
Tests should run fine as though the test collector gem wasn't installed

**Actual**
- If `require "buildkite/test_collector"` is present, but `Buildkite::TestCollector.configure(hook: :minitest)` is not, we get `uninitialized constant Buildkite::TestCollector::MinitestPlugin`

- If neither of the above are present, but the gem is in the Gemfile, we get: `uninitialized constant Minitest::Buildkite`

**Fix**
This just adds an additional constant check in the `Minitest` plugin, and also adds a test to attempt to run Minitest and assert that it runs without raising an exception.

### DIFF
--- a/lib/minitest/buildkite_collector_plugin.rb
+++ b/lib/minitest/buildkite_collector_plugin.rb
@@ -1,6 +1,6 @@
 module Minitest
   def self.plugin_buildkite_collector_init(options)
-    if Buildkite::TestCollector.respond_to?(:uploader)
+    if defined?(Buildkite::TestCollector::MinitestPlugin) && Buildkite::TestCollector.respond_to?(:uploader)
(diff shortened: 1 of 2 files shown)


swebb:
Thanks for your contribution @paulca. I need to check some things internally but should be able to merge soon.

swebb (approved):

paulca:
👍🏻

PR #126, opened 2022-06-20
Fix Minitest crashing without loading minitest lib hook

ChrisBr:
The library hooks are loaded as part of the configure call.
However, we don't always want to run test analytics and therefore
only call configure on certain pipelines, builds etc.
As Minitest automatically discovers plugins, this will crash with a
NameError.


ChrisBr:
> Thanks for your contribution @ChrisBr. Has this issue been resolved by https://github.com/buildkite/test-collector-ruby/pull/125?

🤦‍♂️ I should've checked the open PRs first. Let me check this! Are you planning to release a new version anytime soon?

Issue #122, opened 2022-06-14
ArgumentError, "header #{key} has field value #{value.inspect}, this cannot include CR/LF"


rdh:
I'm getting the following exception after naively copying the API token from the website into GitHub secrets and then setting in ENV for my GitHub action...

```
An error occurred in a `before(:suite)` hook.
Failure/Error: raise ArgumentError, "header #{key} has field value #{value.inspect}, this cannot include CR/LF"
ArgumentError:
  header Authorization has field value "Token token=\"***\n\n\"", this cannot include CR/LF
# /opt/hostedtoolcache/Ruby/2.7.6/x64/lib/ruby/2.7.0/net/http/header.rb:23:in `block in initialize_http_header'
# /opt/hostedtoolcache/Ruby/2.7.6/x64/lib/ruby/2.7.0/net/http/header.rb:16:in `each'
# /opt/hostedtoolcache/Ruby/2.7.6/x64/lib/ruby/2.7.0/net/http/header.rb:16:in `initialize_http_header'
# /opt/hostedtoolcache/Ruby/2.7.6/x64/lib/ruby/2.7.0/net/http/generic_request.rb:46:in `initialize'
# /opt/hostedtoolcache/Ruby/2.7.6/x64/lib/ruby/2.7.0/net/http/request.rb:15:in `initialize'
# /home/runner/work/masterclass/masterclass/backend/vendor/bundle/ruby/2.7.0/gems/buildkite-test_collector-1.0.1/lib/buildkite/test_collector/http_client.rb:19:in `new'
# /home/runner/work/masterclass/masterclass/backend/vendor/bundle/ruby/2.7.0/gems/buildkite-test_collector-1.0.1/lib/buildkite/test_collector/http_client.rb:19:in `post'
# /home/runner/work/masterclass/masterclass/backend/vendor/bundle/ruby/2.7.0/gems/buildkite-test_collector-1.0.1/lib/buildkite/test_collector/uploader.rb:[42](https://github.com/yankaindustries/masterclass/runs/6871164066?check_suite_focus=true#step:7:43):in `configure'
# /home/runner/work/masterclass/masterclass/backend/vendor/bundle/ruby/2.7.0/gems/buildkite-test_collector-1.0.1/lib/buildkite/test_collector/library_hooks/rspec.rb:16:in `block (2 levels) in <top (required)>'
```

I was able to work around this by manually pulling the token from ENV, stripping it, and passing to Buildkite...
```
[...]

PR #127, opened 2022-06-21
Remove whitespace characters from ANALYTICS_TOKEN to avoid multi-line headers

gchan:
Resolves https://github.com/buildkite/test-collector-ruby/issues/122

Ruby 2.5 removed support for multiline headers
https://github.com/ruby/ruby/commit/0078e40115fd53e2cafd4e8cfb6ff5fd7f87c3db

There shouldn't be any new line characters, carriage returns or similar
characters in the BUILDKITE_ANALYTICS_TOKEN environment variable. So we
should remove any instances of leading or trailing whitespace characters
from the token to avoid the error raised in Ruby net/http.

Internal reference:
PIE-927

### DIFF
--- a/lib/buildkite/test_collector.rb
+++ b/lib/buildkite/test_collector.rb
@@ -23,7 +23,7 @@ module Buildkite
     end
 
     def self.configure(hook:, token: nil, url: nil, debug_enabled: false, tracing_enabled: true)
-      self.api_token = token || ENV["BUILDKITE_ANALYTICS_TOKEN"]
+      self.api_token = (token || ENV["BUILDKITE_ANALYTICS_TOKEN"])&.strip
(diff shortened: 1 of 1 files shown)


lib/buildkite/test_collector.rb
```diff
@@ -23,7 +23,7 @@ class << self
     end
 
     def self.configure(hook:, token: nil, url: nil, debug_enabled: false, tracing_enabled: true)
-      self.api_token = token || ENV["BUILDKITE_ANALYTICS_TOKEN"]
+      self.api_token = token || ENV["BUILDKITE_ANALYTICS_TOKEN"]&.strip
```
JuanitoFatas, line 26: We probably want to strip both token passed from the `configure` call and `ENV`? 🤔

PR #128, opened 2022-06-21
Make ruby collector work better for non-web app

JuanitoFatas:
This Pull Request explicitly doing all `require` in the entry point `test_collector.rb`.

For Minitest plugin (file named with `_plugin` suffix) in non-web app, we cant be sure our Collector is `require`’d before Minitest finds the plugin, so need to `require` tracer.

### DIFF
--- a/lib/buildkite/test_collector.rb
+++ b/lib/buildkite/test_collector.rb
@@ -1,16 +1,19 @@
 # frozen_string_literal: true
 
+module Buildkite
+  module TestCollector
+  end
+end
+
 require "timeout"
 require "tmpdir"
 
+require_relative "test_collector/error"
 require_relative "test_collector/version"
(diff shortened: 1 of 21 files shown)


swebb:
I understand why you need to require the tracer inside the plugin, but why do you need/want to move all of the requires to one location? Can't you keep each require closer to where it's used. I'm worried with this approach we won't be removing required libraries/files as they become obsolete.

swebb (approved):
Spoke to @JuanitoFatas about this. Happy to ship with it as-is. Having them in one place doesn't have much of a cost, and putting them in only the files where they are used has proved time consuming. Every failure requires re-building the rails app (for testing) and it takes a long time. We can't a smaller app because it won't provide the same level of coverage.

Issue #131, opened 2022-06-28
Minitest plugin's `before_teardown` can have unexpected interactions with tests


davidstosik:
## What?

We tried using `buildkite-test_collector` and noticed it broke some expectations in our test suite, so we had to temporarily turn it off until we can figure out a workaround or a fix.

## How?

In our test suite, some tests expect a call to be made to `SecureRandom.uuid`, like this:

```rb
SecureRandom.expects(:uuid).returns("123")
```

(We're using the [Mocha](https://github.com/freerange/mocha) mocking/stubbing library and its [`#expects` method](https://github.com/freerange/mocha/blob/58653dbd5eee6921b176e27311936c3dc936dbff/lib/mocha/object_methods.rb#L43-L71).)

The tests used to work but as soon as we started using `buildkite-test_collector`, the tests kept failing, announcing that the expectation was not fulfilled anymore:

```
unexpected invocation: SecureRandom.uuid()
unsatisfied expectations:
- expected exactly once, invoked twice: SecureRandom.uuid(any_parameters)
```

(We also have another one of these unexpected invocations on `Process.clock_gettime(1, :float_second)`.)

Now I know that this kind of expectation is not testing best practice and we have some work to do to educate and figure out better ways, but I think `buildkite-test_collector` is introducing unexpected behaviour, and it would be nice if we could find a way to fix it. 

Taking a quick look at the MiniTest plugin implementation, I noticed the use of `before_teardown`:
https://github.com/buildkite/test-collector-ruby/blob/36f6eee5f988af594f8ebb4a70e8b6dc76596f4d/lib/buildkite/test_collector/minitest_plugin.rb#L15

This gets this to run before all our teardowns, and likely before Mocha's teardown as well.

I'm not sure whether it's possible, but would like to ask: would it still work if this was an `after_teardown` (instead of a `before_teardown`)?

It would kinda make more sense to me too, as running the test collector before teardown means we'll ignore all things that happen in teardown, even at a test file level.

```rb
class SampleTest < Minitest::Test
  def setup
    puts "I'm doing something in #setup, and this will be collected for test analytics."
  end

  def teardown
    puts "I'm doing something in #teardown, and this will NOT be collected for test analytics."
  end

  def sample_test
    puts "This is a sample test"
    assert true
  end
end
```

---

[...]

PR #132, opened 2022-06-28
Use the private reference trick to get time instead

JuanitoFatas:
Avoid breaking test suite that stubs `Process.clock_gettime` by replacing calling `Concurrent.monotonic_time` with a custom `MonotonicTime.call`. 

The fix was suggested by @ChrisBR at https://github.com/buildkite/test-collector-ruby/issues/131#issuecomment-1168343196 (thanks!)

### DIFF
--- a/lib/buildkite/test_collector/tracer.rb
+++ b/lib/buildkite/test_collector/tracer.rb
@@ -4,6 +4,13 @@ require "active_support/core_ext/hash/indifferent_access"
 
 module Buildkite::TestCollector
   class Tracer
+    # https://github.com/buildkite/test-collector-ruby/issues/131
+    class MonotonicTime
+      def self.call
+        Process.method(:clock_gettime).call Process::CLOCK_MONOTONIC
+      end
+    end
+
     class Span
       attr_accessor :section, :start_at, :end_at, :detail, :children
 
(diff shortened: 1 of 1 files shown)


ChrisBr (approved):

lib/buildkite/test_collector/tracer.rb
```diff
@@ -4,6 +4,13 @@
 
 module Buildkite::TestCollector
   class Tracer
+    # https://github.com/buildkite/test-collector-ruby/issues/131
+    class MonotonicTime
+      def self.call
+        Process.method(:clock_gettime).call Process::CLOCK_MONOTONIC
```
swebb, line 10: I'm a bit nervous about this. Have you tested if it works when Process.clock_gettime is stubbed? In the [ruby-timeout implementation](https://github.com/ruby/timeout/pull/18/files#diff-8ad4c142bc670fe498191f02758e39480e7b06e3ff80be9abfc1d73337d35a83R140) they assign the reference to a constant:

image

That constant will be evaluated & stored when the class is defined. If anything mocks the Process.clock_gettime method in future it won't impact the constant and the original method can be called.

In your implementation I think a mocking library would replace `Process.method(:clock_gettime)` and it wouldn't work as desired (although without testing I'm honestly not sure).

JuanitoFatas: Yea I tested the `Process.method(:clock_gettime)` will be mocked and changed back to store in constant instead c70c9a7. Thanks!

swebb (approved):

Issue #135, opened 2022-07-02
Branch misreported under GitHub Actions (or dashboard bug?)


rdh:
When filtering the dashboard to just show `master` branch, it displays no results...
![master](https://user-images.githubusercontent.com/306352/177012950-2c796546-d3a5-406b-8e92-4e0f2bee220b.png)

...we have builds running on `master` but the branch is reported as `refs/heads/master`...
![runs](https://user-images.githubusercontent.com/306352/177012968-ccdf69a2-92fe-4bac-8baa-5da2c560f913.png)

support@buildkite.com states

> The reason for this is because the test have branch environment set as refs/heads/master instead of head as shown below
> Because of this when you filter by master as branch name it did not show any results.

 Perhaps we can just set  BUILDKITE_ANALYTICS_BRANCH to GITHUB_REF with  "ref/heads" prefix removed as a workaround?

gchan:
Hi @rdh 

Thanks for reporting this issue. It looks we should be using `GITHUB_REF_NAME` instead which excludes the prefix

From https://docs.github.com/en/actions/learn-github-actions/environment-variables

![Screenshot from 2022-07-06 11-09-08](https://user-images.githubusercontent.com/1182969/177431522-4e733bd6-12d3-4c91-b92f-dd0df969b659.png)

I have a PR to resolve this issue here https://github.com/buildkite/test-collector-ruby/pull/136

Thanks for reporting this to us! 🙌

gchan:
Hi @rdh 

We have pushed a change to the main branch. Would you be able to test the change in Github Actions before we make a new gem release?

Thanks!

rdh:
Thanks for the quick fix!  We're on company holiday this week and have been ordered by the CEO to "not work".  I'll throw in a PR tomorrow to grab the latest gem source from GitHub, but may be hurting for an approval until Monday.  Stay tuned...

rdh:
Ok.  Included a post-fix version of the gem, like so...

```
  gem 'buildkite-test_collector', git: 'https://github.com/buildkite/test-collector-ruby.git', ref: '2a28da732521cd9984c91e5e7259b71420d6ddb9'
```

...and am now seeing builds reported as being on `master` and the filter is working properly...

<img width="1293" alt="Screen Shot 2022-07-12 at 11 00 05 AM" src="https://user-images.githubusercontent.com/306352/178561785-1e0af489-174e-492a-8fd7-21d08b3dd997.png">

[...]

PR #136, opened 2022-07-05
Remove branch reference prefix in Github Actions

gchan:
Resolves https://github.com/buildkite/test-collector-ruby/issues/135

We are using the environment GITHUB_REF which includes the prefix
ref/branch which isn't what we want. GITHUB_REF_NAME returns the
branch/tag without the prefix.

GITHUB_REF returns
* refs/heads/main
* refs/pulls/2000/merge

GITHUB_REF_NAME returns
* main
* my-branch-name

Reference https://docs.github.com/en/actions/learn-github-actions/environment-variables


swebb (approved):

JuanitoFatas (approved):
👍🏼

PR #137, opened 2022-07-09
Avoid linters complaining about double quotes

SocalNick:
Copied this to an existing project that runs Rubocop during pre-commit hook and it complained that we shouldn't use double quotes when they aren't needed


gchan (approved):
Thanks for your contribution @SocalNick

PR #145, opened 2022-07-28
Suppress errors when connecting to Buildkite

swebb:
It's important the collector does not raise exceptions and cause test
suites to fail. Some customers have seen exceptions raised when
connecting to Buildkite (not a surprise as networks are complicated and
unreliable).

This PR modifies the connection process to catch all StandardErrors and
log them. Trace data will be lost; however, the customer's build will
not fail which is more important.

We also intend to address the underlying issues; however, this PR is an
important defensive measure to protect customer builds.

### DIFF
--- a/lib/buildkite/test_collector.rb
+++ b/lib/buildkite/test_collector.rb
@@ -106,6 +106,7 @@ module Buildkite
       block.call
     rescue StandardError => e
       logger.error("Buildkite::TestCollector received exception: #{e}")
+      logger.error("Backtrace:\n#{e.backtrace.join("\n")}")
     end
   end
 end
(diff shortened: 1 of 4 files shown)


gchan (approved):
> this PR is an
important defensive measure to protect customer builds.

💯

lib/buildkite/test_collector.rb
```diff
@@ -101,5 +101,11 @@ def self.enable_tracing!
         Buildkite::TestCollector::Uploader.tracer&.backfill(:sql, finish - start, **{ query: payload[:sql] })
       end
     end
+
+    def self.safe(&block)
+      block.call
+    rescue StandardError => e
+      logger.error("Buildkite::TestCollector received exception: #{e}")
```
gchan, line 108: Does this also log the backtrace? If not, would it help with debugging the error?

swebb: Good point. I'll add it in.

PR #147, opened 2022-07-29
Fix logging level from debug to error

### LINKED ISSUE #141: Update logging level to 'error'?
https://github.com/buildkite/test-collector-ruby/blob/1118897f153bd248d741d175b0b9b759b69a0d2f/lib/buildkite/test_collector/socket_connection.rb#L81

Should the logging level be `error` to be consistent with the subsequent lines?

https://github.com/buildkite/test-collector-ruby/blob/1118897f153bd248d741d175b0b9b759b69a0d2f/lib/buildkite/test_collector/socket_connection.rb#L88-L103

gchan:
PIE-983

When an EOFError exception is raised, we want log at the 'error' level.
This is also consistent with other exceptions that are rescued.

Resolves #141

### DIFF
--- a/lib/buildkite/test_collector/socket_connection.rb
+++ b/lib/buildkite/test_collector/socket_connection.rb
@@ -78,7 +78,7 @@ module Buildkite::TestCollector
       rescue EOFError => e
         Buildkite::TestCollector.logger.warn("#{e}")
         if @socket
-          Buildkite::TestCollector.logger.warn("attempting disconnected flow")
+          Buildkite::TestCollector.logger.error("attempting disconnected flow")
           @session.disconnected(self)
           disconnect
         end


swebb (approved):
👍️

kivikakk:
The `EOF` case was explicitly separated out from the `ECONNRESET` case and made a warning in #101, presumably because OpenSSL 3 raises these a lot. This PR makes their behaviour identical again, so it's not clear this consistency is actually more correct.

gchan:
@kivikakk thanks for your feedback! Have you been observing noisy errors as a result of this change? We are open to changing it back to a `warn` if you are experiencing `EOFError` cases!

kivikakk:
Yes, indeed! It makes every test end in failure when running on a system with OpenSSL 3:

image

I experienced this because we upgraded our containers from Ubuntu 21.10 ([EOL'd](https://news.itsfoss.com/ubuntu-21-10-eol/)) to 22.04, which includes OpenSSL 3. The test groups pass or fail as usual, but then right at the end:

```
Finished in 4 minutes 4.2 seconds (files took 7.2 seconds to load)
447 examples, 0 failures
#<Thread:0x00007f2477f767a8 /vendor/bundle/ruby/3.1.0/gems/buildkite-test_collector-1.2.3/lib/buildkite/test_collector/socket_connection.rb:64 run> terminated with exception (report_on_exception is true):
/usr/local/lib/ruby/3.1.0/openssl/buffering.rb:157:in `sysread': SSL_read: unexpected eof while reading (OpenSSL::SSL::SSLError)
	from /usr/local/lib/ruby/3.1.0/openssl/buffering.rb:157:in `readpartial'
	from /vendor/bundle/ruby/3.1.0/gems/buildkite-test_collector-1.2.3/lib/buildkite/test_collector/socket_connection.rb:69:in `block in initialize'
bundler: failed to load command: rspec (/vendor/bundle/ruby/3.1.0/bin/rspec)
/usr/local/lib/ruby/3.1.0/openssl/buffering.rb:157:in `sysread': SSL_read: unexpected eof while reading (OpenSSL::SSL::SSLError)
	from /usr/local/lib/ruby/3.1.0/openssl/buffering.rb:157:in `readpartial'
	from /vendor/bundle/ruby/3.1.0/gems/buildkite-test_collector-1.2.3/lib/buildkite/test_collector/socket_connection.rb:69:in `block in initialize'
```

Our short-term solution for this was to simply fork test-collector-ruby and adjust per https://github.com/radiopaedia/test-collector-ruby/commit/e344e2c4b38d738c46b60f0b83bddaea2a6eb312.

gchan:
@kivikakk the forked changes look good to me! Feel free to open a PR from that branch and we can consider getting it merged. Otherwise I can take a look at opening a PR myself.

My only question at the moment is whether there is a more specific subclass of `OpenSSL::SSL::SSLError` we could rescue instead? My concern is that we could be masking critical errors with a `warn` log. Please correct me if I'm mistaken!

kivikakk:
> My only question at the moment is whether there is a more specific subclass of `OpenSSL::SSL::SSLError` we could rescue instead? My concern is that we could be masking critical errors with a `warn` log. Please correct me if I'm mistaken!

I wish that were the case, but as far as I can tell there isn't. (There's only two subclasses, `OpenSSL::SSL::SSLErrorWaitReadable` and `OpenSSL::SSL::SSLErrorWaitWritable`, and neither apply.) Further, Ruby's OpenSSL implementation only exposes the error [through the message](https://github.com/ruby/openssl/blob/173be6690589120976bd3f8e55eea13eae6aed46/ext/openssl/ossl.c#L268-L310), so the best we could do is something like:

```ruby
    # …
  else
    raise
  end
end
```

I don't have the time or energy to put together a better PR for this right now myself, sorry! It was either a hack like above or just switch off test analytics.

Issue #144, opened 2022-07-26
JSON::ParserError results in a failed build


rdh:
Breaking this issue out from https://github.com/buildkite/test-collector-ruby/issues/138 where two different exception types were being discussed.

Over there, we added some extra logging on a branch to try and diagnose this one... https://github.com/buildkite/test-collector-ruby/issues/138#issuecomment-1191969426

Good news: The issue repro'd at 3:55am PT this morning.

Bad news:  The output is missing the log line reporting the JSON payload, which is what we were hoping to see.  It seems like the exception happened twice, but grepping the full GH Action output for "JSON", there's no sign of a "JSON Data" line.

```
2022-07-26T10:45:33.249275260Z pid=7771 tid=3354740 ERROR: JSON::ParserError: JSON::ParserError
#<Thread:0x0000560bb0507a00 /home/runner/work/masterclass/masterclass/backend/vendor/bundle/ruby/2.7.0/bundler/gems/test-collector-ruby-8074177c0faf/lib/buildkite/test_collector/socket_connection.rb:64 run> terminated with exception (report_on_exception is true):
/home/runner/work/masterclass/masterclass/backend/vendor/bundle/ruby/2.7.0/gems/rspec-mocks-3.10.2/lib/rspec/mocks/message_expectation.rb:143:in `block in and_raise': JSON::ParserError (JSON::ParserError)
	from /home/runner/work/masterclass/masterclass/backend/vendor/bundle/ruby/2.7.0/gems/rspec-mocks-3.10.2/lib/rspec/mocks/message_expectation.rb:694:in `block in call'
	from /home/runner/work/masterclass/masterclass/backend/vendor/bundle/ruby/2.7.0/gems/rspec-mocks-3.10.2/lib/rspec/mocks/message_expectation.rb:693:in `map'
	from /home/runner/work/masterclass/masterclass/backend/vendor/bundle/ruby/2.7.0/gems/rspec-mocks-3.10.2/lib/rspec/mocks/message_expectation.rb:693:in `call'
	from /home/runner/work/masterclass/masterclass/backend/vendor/bundle/ruby/2.7.0/gems/rspec-mocks-3.10.2/lib/rspec/mocks/message_expectation.rb:572:in `invoke_incrementing_actual_calls_by'
[...]

PR #149, opened 2022-08-09
Use a private reference to JSON.parse to prevent it being mocked 

gchan:
PIE-979

It is possible for test suites to mock the parse method of JSON.
For example:

allow(JSON).to receive(:parse).and_raise(JSON::ParserError)

This will cause the test collector library to misbehave.

We can resolve this problem by updating the test collector code to
use a private reference to the JSON method which is immune to mocking.

This solution is similar to
https://github.com/buildkite/test-collector-ruby/pull/132
and https://github.com/ruby/timeout/pull/18/files

Resolves https://github.com/buildkite/test-collector-ruby/issues/144


gchan:
There could be other methods that potentially could also be mocked.

Perhaps we should go through a use a private method reference for all of them? WDYT @swebb @JuanitoFatas

JuanitoFatas (approved):
👍🏼 Same trick [we apply to get time](https://github.com/buildkite/test-collector-ruby/pull/132/)

PR #153, opened 2022-09-15
Fix instrumentation of sleep does not support sleeping forever

JuanitoFatas:
[Kernel#sleep](https://rubyapi.org/3.1/o/kernel#method-i-sleep)

When `sleep` calls without duration `sleep()`, it will sleep forever. We should support this behavior to not break people’s test suite.

### DIFF
--- a/lib/buildkite/test_collector/object.rb
+++ b/lib/buildkite/test_collector/object.rb
@@ -7,7 +7,11 @@ module Buildkite::TestCollector
         tracer = Buildkite::TestCollector::Uploader.tracer
         tracer&.enter("sleep")
 
-        super
+        if duration
+          super(duration)
+        else
+          super()
+        end
       ensure
         tracer&.leave
       end
--- a/spec/test_collector/object_spec.rb
+++ b/spec/test_collector/object_spec.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+require "timeout"
+
+RSpec.describe Buildkite::TestCollector::CI do
+  it "sleep without duration should not error" do
+    Buildkite::TestCollector::Object.configure
+
+    expect do
+      ::Timeout.timeout(0.5) do
+        sleep
+      end
+    end.to raise_error(Timeout::Error)
+  end
+end


spec/test_collector/object_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+RSpec.describe Buildkite::TestCollector::CI do
+  it "sleep without duration should not error" do
+    Buildkite::TestCollector::Object.configure
+
+    expect do
+      ::Timeout.timeout(0.5) do
+        sleep
+      end
+    end.to raise_error(Timeout::Error)
```
JuanitoFatas, line 12: without the fix, this will raise an `ArgumentError` when you call `sleep`:

```
  1) Buildkite::TestCollector::CI sleep without duration should not error
     Failure/Error:
       expect do
         ::Timeout.timeout(0.5) do
           sleep
         end
       end.to raise_error(Timeout::Error)

       expected Timeout::Error, got #<ArgumentError: wrong number of arguments (given 0, expected 1)> with backtrace:
         # ./lib/buildkite/test_collector/object.rb:6:in `sleep'
         # ./spec/test_collector/object_spec.rb:10:in `block (4 levels) in <top (required)>'
         # ./spec/test_collector/object_spec.rb:9:in `block (3 levels) in <top (required)>'
         # ./spec/test_collector/object_spec.rb:8:in `block (2 levels) in <top (required)>'
```

lib/buildkite/test_collector/object.rb
```diff
@@ -7,7 +7,11 @@ def sleep(duration)
         tracer = Buildkite::TestCollector::Uploader.tracer
         tracer&.enter("sleep")
 
-        super
+        if duration
+          super(duration)
+        else
+          super()
+        end
```
gchan, line 14: The new test isn't passing. Do we need to make the `duration` parameter optional? If we do that, can we simply call the `super` method without checking `duration`?

JuanitoFatas: Ooops I had the `duration = nil` locally, pushed a new change now.

gchan: Is there anything we can do to simply this block and avoid nil checking `duration`?

JuanitoFatas: super will pass the nil (`duration` defaults to `nil` from method signature) down which will error 😞 

```
Traceback (most recent call last):
        5: from /Users/hhh/.gem/ruby/2.7.6/bin/irb:23:in `<main>'
        4: from /Users/hhh/.gem/ruby/2.7.6/bin/irb:23:in `load'
        3: from /Users/hhh/.rubies/ruby-2.7.6/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        2: from (irb):1
        1: from (irb):1:in `sleep'
TypeError (can't convert NilClass into time interval)
```

gchan (approved):

PR #155, opened 2022-09-20
Allow earlier Rails versions

noodl:
Changes the minimum supported ActiveSupport version to allow the gem to be used with Rails 4.2 applications. This can be helpful for those using RailsLTS to receive backported security fixes to older versions of Rails.


JuanitoFatas (approved):
Awesome. Thanks for the Pull Request!

JuanitoFatas:
@noodl This is released in v1.2.5. Thanks for your contributions again 💚

PR #160, opened 2022-10-14

gchan:
[kivikakk](https://github.com/kivikakk) pointed on that they have been observing lots of `EOF` errors and that logging with a 'warn' is more appropriate than 'error'. https://github.com/buildkite/test-collector-ruby/pull/147#issuecomment-1250485611

Their team has currently forked this Collector as a workaround and it would be great if they don't have to maintain their own fork.

`OpenSSL::SSL::SSLError` is raised when using OpenSSL 3 so catching this and handling it the same way as EOF errors are hadnled seems appropriate.

`OpenSSL::SSL::SSLError` is quite a broad exception class (and Ruby's OpenSSL implementation only exposes the error [through the message](https://github.com/ruby/openssl/blob/173be6690589120976bd3f8e55eea13eae6aed46/ext/openssl/ossl.c#L268-L310), so the best we could do is something like:

```
    # …
  else
    raise
  end
end
```

### DIFF
--- a/lib/buildkite/test_collector/socket_connection.rb
+++ b/lib/buildkite/test_collector/socket_connection.rb
@@ -75,10 +75,13 @@ module Buildkite::TestCollector
       # These get re-raise from session, we should fail gracefully
       rescue *Buildkite::TestCollector::Session::DISCONNECTED_EXCEPTIONS => e
         Buildkite::TestCollector.logger.error("We could not establish a connection with Buildkite Test Analytics. The error was: #{e.message}. If this is a problem, please contact support.")
-      rescue EOFError => e
+      rescue EOFError, OpenSSL::SSL::SSLError => e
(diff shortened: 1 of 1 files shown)


niceking (approved):
This LGTM and I don't have any other context to provide about this. I guess the only consideration would be if the OpenSSL::SSL::SSLError was triggered for a more generic case that just the EOF error, but I think the loss of accuracy between 'error' and 'warn' log levels is acceptable for unblocking failing jobs!

swebb:
@gchan should this PR be merged?

gchan:
@swebb perhaps there should be a spec before we merge?

cc @KatieWright26

gchan:
We are merging this PR without a test given its perceived value and low risk.

Writing the test for this change wasn't straightforward and this hasn't been a high priority for the Test Analytics team. There are future plans to replace the ActionCable/Websockets mode from this collector so we feel comfortable merging this small change as is.

PR #165, opened 2022-12-13
Cope with the gem being loaded but not configured when using minitest

billhorsman:
This problem was introduced in  2eb068dda0a7cf00fdaa644d3204462b5588f2ad.

In our test pipeline, we include the `test-collector-ruby` gem in our `:test` Gemfile group but only configure it to run on certain builds. So we have code like this:

```rb
# Gemfile
group :test do
  gem "buildkite-test_collector", "~> 1.2.5", require: false
end
```

```rb
# test_helper.rb
if !!ENV["BUILDKITE_ANALYTICS_ENABLED"]
  require "buildkite/test_collector"
  Buildkite::TestCollector.configure(hook: :minitest)
end
```

If `BUILDKITE_ANALYTICS_ENABLED` is not set then we never call `configure` and this piece of code is never run:

```rb
# lib/buildkite/test_collector/library_hooks/minitest.rb
Buildkite::TestCollector.uploader = Buildkite::TestCollector::Uploader
```

So `Buildkite::TestCollector.uploader` is nil and this code falls over:

```rb
if trace = Buildkite::TestCollector.uploader.traces[result.source_location]
  Buildkite::TestCollector.session&.write_result(trace)
end
```

This change just wraps that in a test to see if `Buildkite::TestCollector.uploader` is set.

### DIFF
--- a/lib/buildkite/test_collector/minitest_plugin/reporter.rb
+++ b/lib/buildkite/test_collector/minitest_plugin/reporter.rb
@@ -11,8 +11,10 @@ module Buildkite::TestCollector::MinitestPlugin
     def record(result)
       super
 
-      if trace = Buildkite::TestCollector.uploader.traces[result.source_location]
-        Buildkite::TestCollector.session&.write_result(trace)
+      if Buildkite::TestCollector.uploader
+        if trace = Buildkite::TestCollector.uploader.traces[result.source_location]
+          Buildkite::TestCollector.session&.write_result(trace)
+        end
       end
     end
 


gchan:
Thanks for your contribution @billhorsman !

gchan (approved):

gchan:
This issue is potentially related to https://github.com/buildkite/test-collector-ruby/pull/125

The issue is that the plugin is discovered by Minitest before the collector is ever configured (or not configured).
So there isn't a great way to avoid the Reporter from being added to Minitest (i.e. before we check whether it has been configured).

References 
https://github.com/minitest/minitest#writing-extensions-
http://www.monkeyandcrow.com/blog/reading_ruby_minitest_plugin_system/

gchan:
I had a look at the minitest source code. It looks like the plugin files are loaded and the init methods is called when the `run` method is called https://github.com/minitest/minitest/blob/b5565c0c7ab2ce5ecf306487cab0f2abe5377d55/lib/minitest.rb#L140 🤔

billhorsman:
@gchan Thanks for the quick review and release. I've switched to 1.3.1 and all is working nicely again 🙏🏻

PR #167, opened 2022-12-22
Add ability to write test results to a file

niceking:
### Description

This PR adds the functionality to create a gzipped json file that contains the traces for an rspec test suite. The file is saved to a location specified by a new config param. Perhaps this could also become an environment variable later, but this is just a first cut of this idea, and mostly used for testing. 

It shouldn't affect any current functionality of the gem, as the param defaults to nil. 

### Context

This functionality is added in support of testing [PR #10671](https://github.com/buildkite/buildkite/pull/10671), which adds the ability to ingest test results for analytics from an artifact. This creates the artifact for testing purposes. I haven't added any tests, or looked into minitest support for this reason. 

### Verification

I tried this out locally and verified that the appropriate file was created 

### Deployment

Will do a separate PR for a new release with these changes 

### Rollback

Customers can use old releases if this one doesn't work

### DIFF
--- a/lib/buildkite/test_collector/library_hooks/rspec.rb
+++ b/lib/buildkite/test_collector/library_hooks/rspec.rb
@@ -29,6 +29,18 @@ RSpec.configure do |config|
     trace = Buildkite::TestCollector::RSpecPlugin::Trace.new(example, history: tracer.history)
     Buildkite::TestCollector.uploader.traces[example.id] = trace
   end
+
+  config.after(:suite) do
+    if Buildkite::TestCollector.artifact_path
+      filename = File.join(Buildkite::TestCollector.artifact_path, "buildkite-test-collector-rspec-#{SecureRandom.uuid}.json.gz")
+      data_set = { results: Buildkite::TestCollector.uploader.traces.values.map(&:as_hash) }
+      File.open(filename, "wb") do |f|
+        gz = Zlib::GzipWriter.new(f)
+        gz.write(data_set.to_json)
+        gz.close
+      end
+    end
+  end
 end
 
 Buildkite::TestCollector.enable_tracing!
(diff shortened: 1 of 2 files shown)


lib/buildkite/test_collector/library_hooks/rspec.rb
```diff
@@ -29,6 +29,18 @@
     trace = Buildkite::TestCollector::RSpecPlugin::Trace.new(example, history: tracer.history)
     Buildkite::TestCollector.uploader.traces[example.id] = trace
   end
+
+  config.after(:suite) do
+    if Buildkite::TestCollector.artifact_path
+      filename = "#{Buildkite::TestCollector.artifact_path}/rspec-#{SecureRandom.uuid}.json.gz"
```
swebb, line 35: [minor] This would make it slightly more redundant if someone includes a trailing slash in their artifact_path.

```suggestion
      filename = File.join(Buildkite::TestCollector.artifact_path, "rspec-#{SecureRandom.uuid}.json.gz")
```

swebb (approved):
LGTM 👍🏻

lib/buildkite/test_collector/library_hooks/rspec.rb
swebb: I think the filename should reference the collector rather than rspec. How about `buildkite-test_collector-#{SecureRandom.uuid}.json.gz`. Also, it would be great to include the build id rather than a uuid, but that might be more effort than it's worth.

swebb: [minor] Maybe include a debug log that prints the filename created?

niceking: Oh yeah good shout on the `File.join`. 

Yeah I think the file name should reference the collector too, and also have some sort of build/job level identifying ID in the name. The job/build UUIDs are available in the env, but the tricky bit here is that the artifacts need unique names, and we can't guarantee that a build, or a single job, would only contain one test run, so I've just opted for randomness for now. Basically I want to punt on a proper name for this file when it looks like it'll be getting wider release and not just something I'm mostly generating for test purposes

preflightcookie (approved):

PR #170, opened 2023-01-19
[PIE-1365] Update ruby collector API payload

KatieWright26:
This PR tweaks the collector name passed to the Analytics::API so that we can collect accurate information about which collectors are in use by our customers. It is linked to this [piece of work in linear](https://linear.app/buildkite/issue/PIE-435/store-collector-type-junit-rspec-etc).

This PR also adds instructions for deploying this collector, which is consistent with our other test-collectors.

[PIE-1365](https://linear.app/buildkite/issue/PIE-1365/update-ruby-collector-name-to-be-consistent-with-other-collectors)

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,12 +121,17 @@ Bug reports and pull requests are welcome on GitHub at https://github.com/buildk
 
 ## Deploying
 
-1. Bump the version in `version.rb` and run `bundle` to update the `Gemfile.lock`
-1. Update the CHANGELOG.md with your new version and a description of the changes.
-
-Once your PR with your changes is on `main`:
+1. Bump the version in `version.rb` and run `bundle` to update the `Gemfile.lock`.
+1. Update the CHANGELOG.md with your new version and a description of your changes.
+1. Git tag your changes and push
+```
+git tag v.x.x.x
+git push --tags
+```
+Once your PR is merged to `main`:
 
 1. Run `rake release` from `main` and use the rubygems credentials in 1Password.
+1. Create a [new release in github](https://github.com/buildkite/test-collector-ruby/releases).
 1. Bump the gem in [buildkite/buildkite](https://github.com/buildkite/buildkite).
 
 ## 📜 MIT License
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -119,6 +119,21 @@ See [DESIGN.md](DESIGN.md) for an overview of the design of this gem.
...
+```
+git tag v.x.x.x
+git push --tags
+```
+Once your PR is merged to `main`:
+
+1. Run `rake release` from `main` and use the rubygems credentials in 1Password.
+1. Create a [new release in github](https://github.com/buildkite/test-collector-ruby/releases).
+1. Bump the gem in [buildkite/buildkite](https://github.com/buildkite/buildkite).
+
```
niceking, line 136: Since the audience of this readme is the general public, I'm not sure if any of these instruction really belong here. Contributors outside of Buildkite will not be able to access the 1Password, nor bump the gem in buildkite/buildkite. 

For contributors outside of Buildkite, it looks like we follow the pattern of getting contributors to submit a fork with their changes, which is then approved, merged and then released by a Buildkite developer. Perhaps the readme should be edited to reflect that process? 

This is different from what we do with the other collectors, as we are the maintainers of this one, so also keen to hear your thoughts on this as I'm not across the other repos and I know you're trying to standardise them!

KatieWright26: HMMM our other test-collectors are public facing too but have deploy scripts etc in the READMEs... maybe just removing the line about credentials from this one?

niceking: Yeah I feel like having permissions to do the release is sort of implicit? Perhaps we could just remove the reference to creds being in 1Password, cos I wouldn't know how to go about getting the perms to release the python package for example. 

I also think the line about bumping the gem in buildkite could be generalised to "bump the gem in your application" or something like that. 

While I'm being super picky I think the rubygems terminology here is "release" not "deploy"

KatieWright26: Sweet; changes have been made

PR #171, opened 2023-01-26
[PIE-1387] Send `failure_expanded` from minitest

nprizal:
### Description

There is a bug in minitest collector that causes failure_expanded not to be sent to Test Analytics. (see implementation)

minitest is sending a long failure message as failure_reason.

see an example from my local test:

```
RuntimeError: bang!!! 💥
    /Users/naufanrizal/Repositories/minitest-example/test/minitest/test_example.rb:19:in `explode'
    /Users/naufanrizal/Repositories/minitest-example/test/minitest/test_example.rb:8:in `block (2 levels) in <top (required)>'
```

### Context

We want to limit the length of `failure_reason` that can be stored by Test Analytics in the upcoming changes. Therefore, we need to split the failure messages into a short summary that goes into `failure_reason` and the detail that goes into `failure_expanded`

### Changes
1. send the first line of failure message as `failure_reason`
2. send the `failure_expanded` to Test Analytics by removing the first line of failure message from the first failure to avoid duplicate lines in the Test Analytics UI

### Verification
I tested locally by creating a ruby gem scaffold (`bundle gem`) with `minitest` and adding a local build of `build-test_collector` to the `Gemfile`. I then configured the test collector as per [doc](https://buildkite.com/docs/test-analytics/ruby-collectors#minitest-collector) and ran the minitest by pointing to the local `buildkite` app.  

### Deployment
publish to RubyGems in a separate PR

### Rollback
use old releases if this one doesn't work

### DIFF
--- a/lib/buildkite/test_collector/minitest_plugin/trace.rb
+++ b/lib/buildkite/test_collector/minitest_plugin/trace.rb
@@ -72,7 +72,7 @@ module Buildkite::TestCollector::MinitestPlugin
     end
 
     def failure_reason
-      @failure_reason ||= example.failure&.message
+      @failure_reason ||= example.failure&.message&.split("\n")&.first
     end
 
     def failure_expanded
--- a/lib/buildkite/test_collector/minitest_plugin/trace.rb
+++ b/lib/buildkite/test_collector/minitest_plugin/trace.rb
@@ -15,12 +15,10 @@ module Buildkite::TestCollector::MinitestPlugin
 
     FILE_PATH_REGEX = /^(.*?\.(rb|feature))/
 
-    def initialize(example, history:, failure_reason: nil, failure_expanded: [])
+    def initialize(example, history:)
       @id = SecureRandom.uuid
       @example = example
       @history = history
-      @failure_reason = failure_reason
-      @failure_expanded = failure_expanded
     end
 
     def result
@@ -76,13 +74,16 @@ module Buildkite::TestCollector::MinitestPlugin
     end
 
     def failure_expanded
-      @failure_expanded ||= begin
-        example.failures.map do |failure|
-          {
-            expanded: failure.message,
-            backtrace: failure.backtrace,
-          }
-        end
+      @failure_expanded ||= example.failures.map.with_index do |failure, index|
+        # remove the first line of message from the first failure
+        # to avoid duplicate line in Test Analytics UI
+        messages = failure.message.split("\n")
+        messages = messages[1..] if index.zero?
+
+        {
+          expanded: messages,
+          backtrace: failure.backtrace
+        }
       end
     end
 
--- a/lib/buildkite/test_collector/minitest_plugin/trace.rb
+++ b/lib/buildkite/test_collector/minitest_plugin/trace.rb
@@ -70,14 +70,14 @@ module Buildkite::TestCollector::MinitestPlugin
     end
 
     def failure_reason
-      @failure_reason ||= example.failure&.message&.split("\n")&.first
+      @failure_reason ||= strip_invalid_utf8_chars(example.failure&.message)&.split("\n")&.first
     end
 
     def failure_expanded
       @failure_expanded ||= example.failures.map.with_index do |failure, index|
         # remove the first line of message from the first failure
         # to avoid duplicate line in Test Analytics UI
-        messages = failure.message.split("\n")
+        messages = strip_invalid_utf8_chars(failure.message).split("\n")
         messages = messages[1..] if index.zero?
 
         {
--- a/spec/test_collector/minitest_plugin/trace_spec.rb
+++ b/spec/test_collector/minitest_plugin/trace_spec.rb
@@ -28,7 +28,6 @@ RSpec.describe Buildkite::TestCollector::MinitestPlugin::Trace do
     it 'removes invalid UTF-8 characters from top level values' do
       failure_reason = trace.as_hash[:failure_reason]
 
-      expect(failure_reason).to include('test for invalid character')
       expect(failure_reason).to be_valid_encoding
     end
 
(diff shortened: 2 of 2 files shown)


spec/test_collector/minitest_plugin/trace_spec.rb
```diff
@@ -38,6 +45,12 @@
       expect(history_json).to include('347611.734956')
     end
 
+    it 'sets the failure_expanded' do
+      failure_expanded = trace.as_hash[:failure_expanded]
+      expect(failure_expanded).not_to be_empty
+      expect(failure_expanded[0][:expanded]).not_to include('test for invalid character')
+    end
```
gchan, line 52: Do we also want to add tests to verify:
* `failure_reason` only contains the first line?
* `failure_expanded` contains the remaining lines?

lib/buildkite/test_collector/minitest_plugin/trace.rb
```diff
@@ -72,17 +70,20 @@ def project_dir
...
-          }
-        end
+      @failure_expanded ||= example.failures.map.with_index do |failure, index|
+        # remove the first line of message from the first failure
+        # to avoid duplicate line in Test Analytics UI
+        messages = strip_invalid_utf8_chars(failure.message).split("\n")
+        messages = messages[1..] if index.zero?
+
+        {
+          expanded: messages,
```
gchan, line 84: It looks like we are sending an array of strings rather than a string. Is that the intended behaviour?

nprizal: That is what the rspec collector and other collectors are sending (i.e. array of strings). The UI also expects it to be an array of strings, and sending a string will break it. The documentation doesn't explicitly say anything about the expected type for `expanded` but the example has an array of strings. https://buildkite.com/docs/test-analytics/importing-json#json-test-results-data-reference

gchan: So it was likely an error/bug before this change?

nprizal: I won't say it is a bug in the test collector itself because the documentation for `failure_expanded` doesn't specify the expected type. Maybe it is the UI that needs to be more flexible. 

@meghan-kradolfer since you've been working on this area recently, do you have an opinion on this?


Key | Type | Description
-- | -- | --
failure_expanded | array of hashes | A more detailed explanation of why the test failed

gchan (approved):
LGTM, I've assumed the `failure_expanded` format is compatible with the docs https://buildkite.com/docs/test-analytics/importing-json#json-test-results-data-reference

lib/buildkite/test_collector/minitest_plugin/trace.rb
niceking: I wouldn't say that this is an error/bug, but perhaps a quirk of the rspec test collector being designed first, resulting in our data being quite dictated by the shape and design of rspec. 

For a failing test, rspec generates an output that looks like this:
```
     TypeError:
       no implicit conversion of Array into String
     # ./app/workers/analytics/fetch_artifact_for_ingestion_worker.rb:22:in `initialize'
     # ./app/workers/analytics/fetch_artifact_for_ingestion_worker.rb:22:in `new'
     # ./app/workers/analytics/fetch_artifact_for_ingestion_worker.rb:22:in `perform'
     # ./spec/workers/analytics/fetch_artifact_for_ingestion_worker_spec.rb:89:in `block (3 levels) in <top (required)>'
     # ./spec/support/log.rb:17:in `run'
     # ./spec/support/log.rb:66:in `block (2 levels) in <top (required)>'
     # ./spec/support/database.rb:19:in `block (2 levels) in <top (required)>'
```

The first two lines are the "expanded" and the lines starting with `#` become "backtrace". You can see that the [message lines](https://github.com/buildkite/test-collector-ruby/blob/6420d36f5cdb3e5f7888690803e54c3b452fde4b/lib/buildkite/test_collector/rspec_plugin/reporter.rb#L63) that become the data in "expanded" are an array of strings, so we've just replicated that structure. 

I'm not saying that just doing what rspec does is a good pattern to adopt across all our test collectors in different languages, but seeing as the data shown here is basically log lines, I do think having an array of strings is possibly the most straight forward in terms of formatting on the front end? Maybe its annoying to enforce this array of strings format on the collector side, but it seems cleaner than having to parse and split outputs from a variety of different frameworks.

nprizal: I created a post regarding `failure_expanded` on our Basecamp, it will be good to discuss this issue separately.

gchan (approved):

lib/buildkite/test_collector/minitest_plugin/trace.rb
```diff
@@ -72,17 +70,20 @@ def project_dir
     end
 
     def failure_reason
-      @failure_reason ||= example.failure&.message
+      @failure_reason ||= strip_invalid_utf8_chars(example.failure&.message)&.split("\n")&.first
```
niceking, line 73: This is a bit of a nitpick so feel free to ignore, but there's a lot of lonely operators going on in this line, I dunno if its worth refactoring tho 🤷‍♀️

lib/buildkite/test_collector/minitest_plugin/trace.rb
```diff
@@ -72,17 +70,20 @@ def project_dir
...
-          {
-            expanded: failure.message,
-            backtrace: failure.backtrace,
-          }
-        end
+      @failure_expanded ||= example.failures.map.with_index do |failure, index|
+        # remove the first line of message from the first failure
+        # to avoid duplicate line in Test Analytics UI
+        messages = strip_invalid_utf8_chars(failure.message).split("\n")
+        messages = messages[1..] if index.zero?
```
niceking, line 81: I don't understand why we're doing this only for the first failure? Shouldn't we be doing this for all failures? 

I don't know heaps about minitest - but it appears that there could be multiple failures here, much like how rspec does aggregate failures.

spec/test_collector/minitest_plugin/trace_spec.rb
niceking: Should we add some tests for the case where the `failures` array has more than one failure in it? Is it possible to find/generate some legit data for this?

niceking (commented):
I had a little peek into the minitest internals and I think the @example object we're working with is [this one](https://github.com/minitest/minitest/blob/abdde9d03bc8292d14905466d100d2b1e8e844cd/lib/minitest.rb#L528) so it seems like we could have multiple failures for a test, and I'm not sure if that case is well covered by the current implementation

nprizal:
> I had a little peek into the minitest internals and I think the @example object we're working with is [this one](https://github.com/minitest/minitest/blob/abdde9d03bc8292d14905466d100d2b1e8e844cd/lib/minitest.rb#L528) so it seems like we could have multiple failures for a test, and I'm not sure if that case is well covered by the current implementation

> I don't understand why we're doing this only for the first failure? Shouldn't we be doing this for all failures?
> I don't know heaps about minitest - but it appears that there could be multiple failures here, much like how rspec does aggregate failures.

@niceking Thank you for the feedback! I have looked at the `minitest` implementation, and yes, as you said one a test can have multiple failures. But unlike `rspec`, `minitest` doesn't have a "summary" for multiple failures. The `example` object has a `failure` method that returns the first failure from `@failures` attributes (see [here](https://github.com/minitest/minitest/blob/abdde9d03bc8292d14905466d100d2b1e8e844cd/lib/minitest.rb#L427))

In this PR, I'm getting the `failure_reason` from the first line of `example.failure.message` which is equal to `example.failures[0].message`. Therefore I only remove the first line from the first failure.

niceking:
Ahhh yup I get what you're saying now :) So because `example.failure.message` and `example.failures[0].message` are the same, we are skipping the first line for the first failure so we don't show the same thing twice! Cool! Thanks for clearing that up!!!

niceking (approved):

spec/test_collector/minitest_plugin/trace_spec.rb
```diff
@@ -1,12 +1,17 @@
...
 RSpec.describe Buildkite::TestCollector::MinitestPlugin::Trace do
-  subject(:trace) { Buildkite::TestCollector::MinitestPlugin::Trace.new(result, history: history) }
-  let(:result) { double("Result", name: "test_it_passes", test_it_passes: nil, result_code: 'F', failure: failure, failures: [failure]) }
-  let(:failure) { double("Failure", message: message, backtrace: backtrace)}
-  let(:message) { "test for invalid character'\xC8'\n    Expected: true\n    Actual: false" }
+  subject(:trace) { Buildkite::TestCollector::MinitestPlugin::Trace.new(example, history: history) }
+  let(:example) { double("Minitest::Test", name: "test_it_passes", test_it_passes: nil, result_code: 'F', failure: failure, failures: [failure, another_failure]) }
+
+  # failure is either Minitest::Assertion or Minitest::UnexpectedError object. 
+  # ref: https://github.com/minitest/minitest/blob/master/lib/minitest/test.rb#L198
```
niceking, line 11: If you push 'y' while on this page it'll generate a link to this line of code that is pegged to the commit sha so it won't change if the file changes, i.e.

https://github.com/minitest/minitest/blob/0984e29995a5c0f4dcf3c185442bcb4f493ed5e3/lib/minitest/test.rb#L198

niceking (approved):
So good and so thorough!!!!!

gchan (approved):

PR #174, opened 2023-02-24
Batch Upload API requests for rspec

niceking:
### Description

This is the first part of [PIE-1436](https://linear.app/buildkite/issue/PIE-1436/drop-support-for-actioncableuploadchannel-from-the-ruby-collector), which will drop ActionCable support from test-collector-ruby. This removes ActionCable support for test suites configured to use RSpec. 

Use of the SocketSession (websocket implementation) is removed from the RSpec config and replaced with a new Session that sends requests in batches via the Upload API. The batch size is configured via a new env variable, `BUILDKITE_ANALYTICS_UPLOAD_BATCH_SIZE`, which defaults to 500. This is configurable mainly to allow ease of testing. 

The implementation of this leverages RSpec [custom formatters](https://rspec.info/documentation/3.9/rspec-core/RSpec/Core/Formatters.html). After every test, we have defined a custom handler called `handle_example`. The method takes the id of the example which has just executed, and stores it in a buffer. Once this buffer reaches the batch size, the trace data associated with the example ids is uploaded to Buildkite via the Upload API. 

After the test suite has finished executing, the `dump_summary` method is called. This sends any outstanding examples to the Upload API. 

The post requests to Upload API are each done in a separate thread, so that sending large payloads of data do not hold up the execution of the test suite. We retry the Upload API request up to 3 times, if the errors we received look to be network related, rather than a result of not being configured correctly, or the particular suite being over its rate limit. 

At the end of the test suite, the test collector will wait up to 60 seconds for any last Upload API requests to finish, before exiting the test suite. 

### Verification & Deployment

This change can be tested locally by checking out this branch and bundling buildkite/buildkite with this checked out repository, then running `BUILDKITE_ANALYTICS_RSPEC_TOKEN=xx-local-analytics-key BUILDKITE_ANALYTICS_UPLOAD_BATCH_SIZE=5 BUILDKITE_BUILD_ID="1" rspec spec/models/webhook` or similar in the buildkite/buildkite repository. 

After merge, my plan is to do another PR to do a pre-release of this change. Then I will create a branch on buildkite/buildkite to run some scheduled builds with the new release, and monitor for any errors or performance degradations. Typically, an rspec job in Buildkite has 300-600 tests, so we'll be creating 1-2 uploads per rspec job. 

Minitest to come after this!

### DIFF
--- a/lib/buildkite/test_collector.rb
+++ b/lib/buildkite/test_collector.rb
@@ -31,6 +31,7 @@ require_relative "test_collector/object"
 require_relative "test_collector/tracer"
 require_relative "test_collector/socket_connection"
 require_relative "test_collector/socket_session"
+require_relative "test_collector/session"
 
 module Buildkite
   module TestCollector
--- a/lib/buildkite/test_collector.rb
+++ b/lib/buildkite/test_collector.rb
@@ -35,6 +35,7 @@ require_relative "test_collector/socket_session"
 module Buildkite
   module TestCollector
     DEFAULT_URL = "https://analytics-api.buildkite.com/v1/uploads"
+    DEFAULT_UPLOAD_BATCH_SIZE = 500
 
     class << self
       attr_accessor :api_token
@@ -45,6 +46,7 @@ module Buildkite
       attr_accessor :tracing_enabled
       attr_accessor :artifact_path
       attr_accessor :env
+      attr_accessor :batch_size
     end
 
     def self.configure(hook:, token: nil, url: nil, debug_enabled: false, tracing_enabled: true, artifact_path: nil, env: {})
@@ -54,7 +56,7 @@ module Buildkite
       self.tracing_enabled = tracing_enabled
       self.artifact_path = artifact_path
       self.env = env
-
+      self.batch_size = ENV.fetch("BUILDKITE_ANALYTICS_UPLOAD_BATCH_SIZE") { DEFAULT_UPLOAD_BATCH_SIZE }.to_i
       self.hook_into(hook)
     end
 
--- a/lib/buildkite/test_collector/http_client.rb
+++ b/lib/buildkite/test_collector/http_client.rb
@@ -28,6 +28,28 @@ module Buildkite::TestCollector
       http.request(contact)
     end
 
+    def post_json(data)
+      contact_uri = URI.parse(url)
+
+      http = Net::HTTP.new(contact_uri.host, contact_uri.port)
+      http.use_ssl = contact_uri.scheme == "https"
+
+      contact = Net::HTTP::Post.new(contact_uri.path, {
+        "Authorization" => authorization_header,
+        "Content-Type" => "application/json",
+      })
+
+      data_set = data.map(&:as_hash)
+
+      contact.body = {
+        run_env: Buildkite::TestCollector::CI.env,
+        format: "json",
+        data: data_set
+      }.to_json
+
+      http.request(contact)
+    end
+
     private
 
     attr :url
--- a/lib/buildkite/test_collector/library_hooks/rspec.rb
+++ b/lib/buildkite/test_collector/library_hooks/rspec.rb
@@ -11,8 +11,6 @@ Buildkite::TestCollector.uploader = Buildkite::TestCollector::Uploader
 RSpec.configure do |config|
   config.before(:suite) do
     config.add_formatter Buildkite::TestCollector::RSpecPlugin::Reporter
-
-    Buildkite::TestCollector.safe { Buildkite::TestCollector::Uploader.configure }
   end
 
   config.around(:each) do |example|
--- a/lib/buildkite/test_collector/rspec_plugin/reporter.rb
+++ b/lib/buildkite/test_collector/rspec_plugin/reporter.rb
@@ -7,6 +7,7 @@ module Buildkite::TestCollector::RSpecPlugin
     attr_reader :output
 
     def initialize(output)
+      Buildkite::TestCollector.session = Buildkite::TestCollector::Session.new
       @output = output
     end
 
@@ -19,21 +20,13 @@ module Buildkite::TestCollector::RSpecPlugin
         if example.execution_result.status == :failed
           trace.failure_reason, trace.failure_expanded = failure_info(notification)
         end
-        Buildkite::TestCollector.session&.write_result(trace)
+        Buildkite::TestCollector.session.add_example_to_send_queue(example.id)
       end
     end
 
-    def dump_summary(notification)
-      if Buildkite::TestCollector.session.present?
-        examples_count = {
-          examples: notification.examples.count,
-          failed: notification.failed_examples.count,
-          pending: notification.pending_examples.count,
-          errors_outside_examples: notification.errors_outside_of_examples_count
-        }
-
-        Buildkite::TestCollector.session.close(examples_count)
-      end
+    def dump_summary(_notification)
+      Buildkite::TestCollector.session.send_remaining_data
+      Buildkite::TestCollector.session.close
     end
 
     alias_method :example_passed, :handle_example
--- a/lib/buildkite/test_collector/session.rb
+++ b/lib/buildkite/test_collector/session.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+module Buildkite::TestCollector
+  class Session
+    UPLOAD_THREAD_TIMEOUT = 60
+    UPLOAD_SESSION_TIMEOUT = 60
+    UPLOAD_API_MAX_RESULTS = 5000
+
+    def initialize
+      @send_queue_ids = []
+      @upload_threads = []
+    end
+
+    def add_example_to_send_queue(id)
+      @send_queue_ids << id
+
+      if @send_queue_ids.size >= Buildkite::TestCollector.batch_size
+        send_ids = @send_queue_ids.shift(Buildkite::TestCollector.batch_size)
+        upload_data(send_ids)
+      end
+    end
+
+    def send_remaining_data
+      return if @send_queue_ids.empty?
+
+      upload_data(@send_queue_ids)
+    end
+
+    def close
+      # There are two thread joins here, because the inner join will wait up to
+      # UPLOAD_THREAD_TIMEOUT seconds PER thread that is uploading data, i.e.
+      # n_threads x UPLOAD_THREAD_TIMEOUT latency if Buildkite happens to be
+      # down. By wrapping that in an outer thread join with the
+      # UPLOAD_SESSION_TIMEOUT, we ensure that we only wait a max of
+      # UPLOAD_SESSION_TIMEOUT seconds before the session exits.
+      Thread.new do
+        @upload_threads.each { |t| t.join(UPLOAD_THREAD_TIMEOUT) }
+      end.join(UPLOAD_SESSION_TIMEOUT)
+
+      @upload_threads.each { |t| t&.kill }
+    end
+
+    private
+
+    def upload_data(ids)
+      data = Buildkite::TestCollector.uploader.traces.values_at(*ids).compact
+
+      # we do this in batches of UPLOAD_API_MAX_RESULTS in case the number of
+      # results exceeds this due to a bug, or user error in configuring the
+      # batch size
+      data.each_slice(UPLOAD_API_MAX_RESULTS) do |batch|
+        new_thread = Buildkite::TestCollector::Uploader.upload(batch)
+        @upload_threads << new_thread if new_thread
+      end
+    end
+  end
+end
--- a/lib/buildkite/test_collector/uploader.rb
+++ b/lib/buildkite/test_collector/uploader.rb
@@ -2,6 +2,8 @@
 
 module Buildkite::TestCollector
   class Uploader
+    MAX_UPLOAD_ATTEMPTS = 3
+
     def self.traces
       @traces ||= {}
     end
@@ -10,6 +12,14 @@ module Buildkite::TestCollector
       URI::InvalidURIError,
       Net::HTTPBadResponse,
       Net::HTTPHeaderSyntaxError,
+      Net::ReadTimeout,
+      Net::OpenTimeout,
+      OpenSSL::SSL::SSLError,
+      OpenSSL::SSL::SSLErrorWaitReadable,
+      EOFError
+    ]
+
+    RETRYABLE_UPLOAD_ERRORS = [
       Net::ReadTimeout,
       Net::OpenTimeout,
       OpenSSL::SSL::SSLError,
@@ -53,6 +63,23 @@ module Buildkite::TestCollector
 
     def self.tracer
       Thread.current[:_buildkite_tracer]
+    end
+
+    def self.upload(data)
+      return false unless Buildkite::TestCollector.api_token
+
+      http = Buildkite::TestCollector::HTTPClient.new(Buildkite::TestCollector.url)
+
+      Thread.new do
+        response = begin
+          upload_attempts ||= 0
+          http.post_json(data)
+        rescue *Buildkite::TestCollector::Uploader::RETRYABLE_UPLOAD_ERRORS => e
+          if (upload_attempts += 1) < MAX_UPLOAD_ATTEMPTS
+            retry
+          end
+        end
+      end
     end
   end
 end
--- a/spec/test_collector/session_spec.rb
+++ b/spec/test_collector/session_spec.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+RSpec.describe Buildkite::TestCollector::Session do
+  subject { described_class.new }
+  let(:data) do
+    {
+      "test-1": "test-1-data",
+      "test-2": "test-2-data",
+      "test-3": "test-3-data",
+      "test-4": "test-4-data",
+      "test-5": "test-5-data",
+    }
+  end
+
+  before do
+    Buildkite::TestCollector.uploader = Buildkite::TestCollector::Uploader
+    data.each { |k, v| Buildkite::TestCollector.uploader.traces[k] = v }
+    Buildkite::TestCollector.batch_size = 5
+  end
+
+  describe "#add_example_to_send_queue" do
+    it "adds the example and sends only when batch size is reached" do
+      expect(Buildkite::TestCollector::Uploader).to receive(:upload).once.with(data.values)
+
+      subject.add_example_to_send_queue(:"test-1")
+      subject.add_example_to_send_queue(:"test-2")
+      subject.add_example_to_send_queue(:"test-3")
+      subject.add_example_to_send_queue(:"test-4")
+      subject.add_example_to_send_queue(:"test-5")
+    end
+  end
+
+  describe "#send_remaining_data" do
+    it "sends through remaining examples even when batch size is not reach" do
+      expect(Buildkite::TestCollector::Uploader).to receive(:upload).once.with(data.reject { |k,v| k == :"test-5"}.values)
+
+      subject.add_example_to_send_queue(:"test-1")
+      subject.add_example_to_send_queue(:"test-2")
+      subject.add_example_to_send_queue(:"test-3")
+      subject.add_example_to_send_queue(:"test-4")
+
+      subject.send_remaining_data
+    end
+
+    it "limits upload batch size to UPLOAD_API_MAX_RESULTS" do
+      stub_const("Buildkite::TestCollector::Session::UPLOAD_API_MAX_RESULTS", 1)
+
+      expect(Buildkite::TestCollector::Uploader).to receive(:upload).once.with(["test-1-data"])
+      expect(Buildkite::TestCollector::Uploader).to receive(:upload).once.with(["test-2-data"])
+
+      subject.add_example_to_send_queue(:"test-1")
+      subject.add_example_to_send_queue(:"test-2")
+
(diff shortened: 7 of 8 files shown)


lib/buildkite/test_collector/library_hooks/rspec.rb
```diff
@@ -12,7 +12,8 @@
   config.before(:suite) do
     config.add_formatter Buildkite::TestCollector::RSpecPlugin::Reporter
 
-    Buildkite::TestCollector.safe { Buildkite::TestCollector::Uploader.configure }
```
niceking, line 15: This removes initialization of the session, and the socket

lib/buildkite/test_collector/library_hooks/rspec.rb
```diff
@@ -12,7 +12,8 @@
...
+    Thread.current[:_buildkite_send_queue] = []
+    Thread.current[:_buildkite_upload_threads] = []
```
niceking, line 16: I'm just putting these variable in the thread local variable space because I haven't created a new session class to store stuff like this. But all the code that runs in these various config blocks are in the main thread, and so it is shared

lib/buildkite/test_collector/library_hooks/rspec.rb
```diff
@@ -28,9 +29,36 @@
...
+
+    if data
+      new_thread = Buildkite::TestCollector::Uploader.upload(data)
+      Thread.current[:_buildkite_upload_threads] << new_thread if new_thread
+    end
+
+    puts "🫡🫡🫡 waiting up 60s for threads to finish #{Time.now}"
+    Thread.new(Thread.current[:_buildkite_upload_threads]) do |threads|
+      threads.each { |t| t.join(60) }
+    end.join(60)
```
niceking, line 57: I have two sets of joins here, because the inner `threads.each { |t| t.join(60) }` will wait up to 60s PER thread that is uploading data, i.e. n_threads x 60s latency if Buildkite happens to be down. By wrapping that in an outer thread with a 60s timeout, we ensure that we only wait 60s. 

All the threads are still operating in the background though, so none of this stops the data from being sent, it just stops us from waiting for ages at the end of the test run

lib/buildkite/test_collector/library_hooks/rspec.rb
```diff
@@ -28,9 +29,36 @@
...
+    puts "🫡🫡🫡 have finished waiting #{Time.now}"
+
+    Thread.current[:_buildkite_upload_threads].each { |t| t&.kill }
```
niceking, line 60: Safely terminate any threads that haven't finished sending data via the upload api

swebb:
@niceking did you accidentally force push from the wrong branch? This looks like the change set for renaming session to socket session.

swebb:
I had a look at some of the earlier commits. Am I reading it correctly in thinking the uploads will be in batches of 5 executions at a time. That number seems very low to me. I'm worried it will result in lots of uploads (a single run of bk/bk generates 13k executions, so 2.5k uploads). I'm worried some of our queries won't work well with that number of uploads. Also, that will result in 2.5k files per run stored in S3 (I'm not sure if there's any limits/costs there).

swebb:
I'm impressed you managed this with such a small change set 👍🏻

niceking:
@swebb hahaha yup I did a bad rebase as I had changed the parent branch 😓 thank god for git reflog. And yeah, the batch size was just set to 5 for testing and demo purposes

lib/buildkite/test_collector/session.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def close
+      # There are two thread joins here, because the inner join will wait up to
+      # UPLOAD_THREAD_TIMEOUT seconds PER thread that is uploading data, i.e.
+      # n_threads x UPLOAD_THREAD_TIMEOUT latency if Buildkite happens to be
+      # down. By wrapping that in an outer thread join with the
+      # UPLOAD_SESSION_TIMEOUT, we ensure that we only wait a max of
+      # UPLOAD_SESSION_TIMEOUT seconds before the session exits.
+      Thread.new do
+        @upload_threads.each { |t| t.join(UPLOAD_THREAD_TIMEOUT) }
```
blaknite, line 36: My understanding of `join` is that any exceptions from the thread it is called on are now processed in this thread. Is this ok? If so then this approach is the easiest to reason with.

lib/buildkite/test_collector/uploader.rb
```diff
@@ -54,5 +64,23 @@ def self.configure
...
+        response = begin
+          upload_attempts ||= 1
+          http.post_json(data)
+        rescue *Buildkite::TestCollector::Uploader::RETRYABLE_UPLOAD_ERRORS => e
+          if (upload_attempts += 1) > MAX_UPLOAD_ATTEMPTS
+            retry
+          end
+        end
+      end
+
```
blaknite, line 83:
```suggestion
```

blaknite (commented):
So much less code than the websocket version 🙌

I've left a few questions, comments, and suggestions for your contemplation 🧠

lib/buildkite/test_collector.rb
```diff
@@ -54,7 +57,7 @@ def self.configure(hook:, token: nil, url: nil, debug_enabled: false, tracing_en
       self.tracing_enabled = tracing_enabled
       self.artifact_path = artifact_path
       self.env = env
-
+      self.batch_size = ENV["BUILDKITE_ANALYTICS_UPLOAD_BATCH_SIZE"].present? ? ENV["BUILDKITE_ANALYTICS_UPLOAD_BATCH_SIZE"].to_i : DEFAULT_UPLOAD_BATCH_SIZE
```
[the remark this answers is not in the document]
niceking: Will roll the dice with no one defining empty string values 🤠

lib/buildkite/test_collector/session.rb
```diff
@@ -0,0 +1,50 @@
...
+      if @send_queue_ids.size >= Buildkite::TestCollector.batch_size
+        send_ids = @send_queue_ids.shift(Buildkite::TestCollector.batch_size)
+        upload_data(send_ids)
+      end
+    end
+
+    def send_remaining_data
+      return if @send_queue_ids.empty?
+
+      upload_data(@send_queue_ids)
```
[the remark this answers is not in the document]
niceking: Doesn't matter because there's no check on queue size in `upload_data`, it just sends whatever ids its passed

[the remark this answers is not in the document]
niceking: Ooo, I don't think that's an issue unless queue size is bigger than the upload api limit of 5000. As batch size is user configurable, I've changed this method to batch results in 5000, as that's easier than somehow validating the user supplied batch size.

lib/buildkite/test_collector/session.rb
```diff
@@ -0,0 +1,50 @@
...
+        @upload_threads.each { |t| t.join(UPLOAD_THREAD_TIMEOUT) }
+      end.join(UPLOAD_SESSION_TIMEOUT)
+
+      @upload_threads.each { |t| t&.kill }
+    end
+
+    private
+
+    def upload_data(ids)
+      data = ids.map { |id| Buildkite::TestCollector.uploader.traces[id] }
```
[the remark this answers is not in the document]
niceking: We use the trace hash later to make the artifact file:

https://github.com/buildkite/test-collector-ruby/blob/2715713f506977d7e5871e1a6e3ed58b25fa23c8/lib/buildkite/test_collector/library_hooks/rspec.rb#L33-L43

lib/buildkite/test_collector/session.rb
niceking: Mmm yeah, I guess we could wrap this entire block in a begin rescue block and catch StandardError, so whatever exceptions that could get thrown in the thread don't make the rspec job blow up? 

Maybe while we're just testing this on buildkite/buildkite we don't do that so we can see what errors we're getting and deal with them appropriately, but before this gem is released for customer use we put in a protection like that so we don't break any customer builds?

lib/buildkite/test_collector/http_client.rb
```diff
@@ -28,6 +28,28 @@ def post
       http.request(contact)
     end
 
+    def post_json(data)
+      contact_uri = URI.parse(url)
+
+      http = Net::HTTP.new(contact_uri.host, contact_uri.port)
+      http.use_ssl = contact_uri.scheme == "https"
```
KatieWright26, line 35: Is contact_uri just the buildkite URL? I see this is a pattern above, but why we don't just set this to true?

lib/buildkite/test_collector/rspec_plugin/reporter.rb
```diff
@@ -34,6 +35,13 @@ def dump_summary(notification)
 
         Buildkite::TestCollector.session.close(examples_count)
       end
+
+      Buildkite::TestCollector.session.add_example_to_send_queue(example.id)
+    end
+
+    def dump_summary(_)
```
KatieWright26, line 42: I can't find the correct search term for this, why do you pass an underscore as an argument here?

lib/buildkite/test_collector/uploader.rb
```diff
@@ -54,5 +64,22 @@ def self.configure
...
+      return false unless Buildkite::TestCollector.api_token
+
+      http = Buildkite::TestCollector::HTTPClient.new(Buildkite::TestCollector.url)
+
+      Thread.new do
+        response = begin
+          upload_attempts ||= 1
+          http.post_json(data)
+        rescue *Buildkite::TestCollector::Uploader::RETRYABLE_UPLOAD_ERRORS => e
+          if (upload_attempts += 1) > MAX_UPLOAD_ATTEMPTS
```
KatieWright26, line 78: Should this be `(upload_attempts += 1) <= MAX_UPLOAD_ATTEMPTS`?

KatieWright26 (commented):
I think I follow this at a high level but have left some questions around bits I can't figure out myself 🍾

lib/buildkite/test_collector/http_client.rb
niceking: The URL is configurable here:
https://github.com/buildkite/test-collector-ruby/blob/4f2ee8dae18274ecceb1aa3ffadedc1d60591ab4/lib/buildkite/test_collector.rb#L52

So for instance when we use the test collector in dev, the contact uri scheme is not https
https://github.com/buildkite/buildkite/blob/b34312e4b7d643eaeae064cc0b387ab8927872db/spec/spec_helper.rb#L8

lib/buildkite/test_collector/rspec_plugin/reporter.rb
```diff
@@ -7,6 +7,7 @@ class Reporter
     attr_reader :output
 
     def initialize(output)
+      Buildkite::TestCollector.session = Buildkite::TestCollector::Session.new
```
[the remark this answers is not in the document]
gchan: I suspect for `start` to function. We would need to register this formatter class earlier (outside the before :suite block)

https://github.com/buildkite/test-collector-ruby/blob/803fdda039e0aeaeb92bcf89cae1fbccddab2480/lib/buildkite/test_collector/library_hooks/rspec.rb#L11-L14

lib/buildkite/test_collector/session.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def upload_data(ids)
+      data = Buildkite::TestCollector.uploader.traces.values_at(*ids).compact
+
+      # we do this in batches of UPLOAD_API_MAX_RESULTS in case the number of
+      # results exceeds this due to a bug, or user error in configuring the
+      # batch size
+      data.each_slice(UPLOAD_API_MAX_RESULTS) do |batch|
+        new_thread = Buildkite::TestCollector::Uploader.upload(data)
+        @upload_threads << new_thread if new_thread
```
gchan, line 53: My understanding is that `new_thread` may be nil if `Buildkite::TestCollector.api_token` isn't configured.
I wonder it's worth shifting/copying this api_token check here or up to the formatter class? Or does the formatter still need to function if the api_token is absent?

lib/buildkite/test_collector/session.rb
gchan: Based on my reading of the docs for the `join` method, exceptions are not raised if [::abort_on_exception](https://github.com/buildkite/test-collector-ruby/pull/174/Thread.html#method-c-abort_on_exception) is set to false.  And it seems the default value for `abort_on_exception` is `false`.

If you need to catch StandardError, you could consider making the `upload` method handle it itself inside the thread.

As a further note - there may also be other rescues for StandardError further up the call stack as well due to the strategy of not breaking builds if there is an exceptional error in this Ruby Collector.

lib/buildkite/test_collector/session.rb
```diff
@@ -0,0 +1,57 @@
...
+      # UPLOAD_THREAD_TIMEOUT seconds PER thread that is uploading data, i.e.
+      # n_threads x UPLOAD_THREAD_TIMEOUT latency if Buildkite happens to be
+      # down. By wrapping that in an outer thread join with the
+      # UPLOAD_SESSION_TIMEOUT, we ensure that we only wait a max of
+      # UPLOAD_SESSION_TIMEOUT seconds before the session exits.
+      Thread.new do
+        @upload_threads.each { |t| t.join(UPLOAD_THREAD_TIMEOUT) }
+      end.join(UPLOAD_SESSION_TIMEOUT)
+
+      @upload_threads.each { |t| t&.kill }
```
gchan, line 40: What's the reason behind the lonely operator? Do we expect nils in `@upload_threads`?

Nitpick - can we kill the threads if `join` times out and returns nil on line 37?

spec/test_collector/session_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      subject.add_example_to_send_queue(:"test-3")
+      subject.add_example_to_send_queue(:"test-4")
+
+      subject.send_remaining_data
+    end
+
+    it "limits upload batch size to UPLOAD_API_MAX_RESULTS" do
+      stub_const("Buildkite::TestCollector::Session::UPLOAD_API_MAX_RESULTS", 1)
+
+      expect(Buildkite::TestCollector::Uploader).to receive(:upload).twice
```
gchan, line 48: Should we verify the data upload receives each time?

gchan (commented):
This looks great. I like your plan to pre-release this for testing.

I've left a bunch of questions/nits/suggestions as comments.

This is one major question about the duplicate dump_summary method in the reporter (more details in the comment).

Also, should we update/add specs to reporter_spec.rb to test the integration with the new Session class?

lib/buildkite/test_collector/rspec_plugin/reporter.rb
niceking: There's a convention in Ruby where unused variables are prefixed with underscore, or just stored as underscore. As `dump_summary` is an alias of three methods that rspec expects to call with an argument, I've just stored the argument here as `_` to indicate its unused. I could also have done `_notification` or something similar. Using just underscore itself is [slightly magic](https://po-ru.com/2012/09/21/rubys-magic-underscore) cos you can do it for multiple arguments. 

I think we also have a [linter](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/UnusedMethodArgument) for this in buildkite/buildkite but not on this repo.

lib/buildkite/test_collector/uploader.rb
niceking: Ooo good catch!

lib/buildkite/test_collector/rspec_plugin/reporter.rb
niceking: Lol jinx @gchan I didn't see your comment. I can't see where you're seeing two `dump_summary` methods?

lib/buildkite/test_collector/rspec_plugin/reporter.rb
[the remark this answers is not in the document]
niceking: Yeah I'm not sure this is worth the effort

lib/buildkite/test_collector/rspec_plugin/reporter.rb
KatieWright26: Oh yep, I'm familiar with the `_` syntax, but was thrown by the lack of a argument name. +1 to @gchan's suggestion for readability

lib/buildkite/test_collector/rspec_plugin/reporter.rb
```diff
@@ -34,6 +35,13 @@ def dump_summary(notification)
 
         Buildkite::TestCollector.session.close(examples_count)
       end
+
+      Buildkite::TestCollector.session.add_example_to_send_queue(example.id)
+    end
+
+    def dump_summary(_)
+      Buildkite::TestCollector.session.send_remaining_data
+      Buildkite::TestCollector.session.close
```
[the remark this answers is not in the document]
niceking: I think its clearer to keep these two methods like this and know from reading it that we always send the remaining data before we close out the session, less knowledge of rspec internals needed to understand the execution order

lib/buildkite/test_collector/session.rb
niceking: I had the lonely operator in there cos I thought you couldn't kill a thread that was already dead (i.e. the upload has already been sent) but it seems you can call kill on a dead thread still so I guess there's no reason to keep the lonely operator there. It's also safe to kill a thread that has timed out its join

lib/buildkite/test_collector/session.rb
niceking: I think its splitting hairs on whether its more appropriate check the token either here or in the upload method, I put it in the upload method because its closer to where the api request is actually made

spec/test_collector/session_spec.rb
niceking: Maybe? Not sure how easy this is to do tbh, I don't think you can chain a `with` call with two different sets of arguments after a `twice`? Plus we're already verifying the data in the previous test so 🤷‍♀️

lib/buildkite/test_collector/session.rb
niceking: Report on exception defaults to true: https://ruby-doc.org/core-2.5.0/Thread.html#method-c-report_on_exception

niceking: @gchan and I caught up about this 

[abort_on_exception](https://ruby-doc.org/3.0.5/Thread.html#method-c-abort_on_exception) controls whether errors in threads are raised _immediately_ to the main thread or not. This defaults to false.

When `join` is called on a thread that has thrown an exception, that [exception is then re-raised on the main thread](https://ruby-doc.org/3.0.5/Thread.html#class-Thread-label-Exception+handling)

While we are still testing and only using this new version of the gem internally, it's better to have more visibility on the exceptions so we can deal with any errors. Before public release of this gem, wrap this bit of code, and any other risky bits of code in the [safe method](https://github.com/buildkite/test-collector-ruby/blob/4f2ee8dae18274ecceb1aa3ffadedc1d60591ab4/lib/buildkite/test_collector.rb#L109) or similar

spec/test_collector/session_spec.rb
gchan: I think it's as easy as writing two separate expectations?

My thinking behind testing the arguments is to verify there isn't a bug with the incorrect executions being upload each time it is called.

I don't think the previous specs test for the case when the batch size is reached prematurely before `send_remaining_data` is called.

lib/buildkite/test_collector/uploader.rb
niceking: It's pretty hard to write a spec for this because its happening in a separate thread, any mock/stub/spy we can create in rspec won't be visible to this thread because its happening in a separate thread

jasmine-q:
I had a look through this and I think there's some weird stuff going on...probs a bad rebase @niceking? 
like `Buildkite::TestCollector.session&.write_result(trace)` is still in `handle_example` and I thought this was meant to remove the old websocket implementation 😕

niceking:
@jasmine-q yes sorry bad rebase! Have a look now :)

lib/buildkite/test_collector/uploader.rb
```diff
@@ -54,5 +64,22 @@ def self.configure
...
+
+      http = Buildkite::TestCollector::HTTPClient.new(Buildkite::TestCollector.url)
+
+      Thread.new do
+        response = begin
+          upload_attempts ||= 1
+          http.post_json(data)
+        rescue *Buildkite::TestCollector::Uploader::RETRYABLE_UPLOAD_ERRORS => e
+          if (upload_attempts += 1) < MAX_UPLOAD_ATTEMPTS
+            retry
```
gchan, line 79: I figured out how to test threaded code with this one weird trick and discovered this might be off by one. We could resolve this by setting upload_attempts to 0 on line 75

lib/buildkite/test_collector/uploader.rb
```diff
@@ -54,5 +64,22 @@ def self.configure
...
+          end
```
gchan, line 80: I figured out how to test threaded code in rspec with this one weird trick and discovered this might be off by one?

blaknite (approved):

lib/buildkite/test_collector/uploader.rb
gchan: Fixed!

gchan (approved):
LGTM! 🎉 Thanks for fixing those last two minor issues I identified this morning ⚒️

PR #175, opened 2023-02-27
Rename session to socket session

niceking:
This PR is just a precursor to #174, which switches test-collector-ruby to send upload data via the Upload API rather than the ActionCable websocket for RSpec test suites. 

This PR renames the previous `Session` to `SocketSession` as Minitest is still using the websocket implementation. As PR #174 creates a new `Session` object, the rename was pulled out into a separate PR for clarity of reading PR #174


gchan (approved):
LGTM. Thanks for pulling this change out of the other PR!

blaknite (approved):

PR #177, opened 2023-03-06
[PIE-1436] Pre-release Rspec changes

niceking:
Creating a pre-release for the major changes to the Rspec plugin 

Have given it another integration test locally by running `BUILDKITE_ANALYTICS_RSPEC_TOKEN=xx-local-analytics-key BUILDKITE_ANALYTICS_UPLOAD_BATCH_SIZE=2 BUILDKITE_BUILD_ID="hi" rspec spec/models/admin/goto_spec.rb` and verified that all relevant records and S3 files are created.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## v2.0.0.pre
+
+- Major change: RSpec plugin to use HTTP Upload API instead of websocket connection to send test data #174 #175 - @niceking
+- `identifier` field removed from trace #176 - @amybiyuliu
+
 ## v1.5.0
 
 - Send `failure_expanded` from minitest #171 - @nprizal 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    buildkite-test_collector (1.5.0)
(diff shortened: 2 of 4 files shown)


KatieWright26 (approved):
LGTM

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## v2.0.0.pre
+
+- Major change: RSpec plugin to use HTTP Upload API instead of websocket connection to send test data #174 #175 - @niceking
+- `identifier` field removed from trace #176 - @amybiyuliu
```
gchan, line 6: Missing #160

```suggestion
- `identifier` field removed from trace #176 - @amybiyuliu
- Only warn on EOF errors and also catch SSLErrors #160 - @gchan
```

niceking: The #160 PR went in like a month ago and there's been 5 releases with that change since then? You're welcome to open another PR to put #160 in the changelog for v1.3.0 but I'm not going to include it here because it isn't a change between v1.5.0 and v2.0.0

gchan: @niceking I don't think #160 ever made it into a release? It was merged after the release of v1.5.0!

PR #179, opened 2023-03-29
Remove websocket and debug functionality

niceking:
Time for the ol' ruby test collector to get a 💇‍♀️ 

I've removed three things:
1. All the code that established the websocket connection, and the session that managed sending data through the websocket 
2. Debug mode - basically the only thing that we used the logging for was for the socket session because it was so complicated. There isn't really anything to log anymore. Hence I have removed it. Note that this is a breaking change to the `Buildkite::TestCollector.configure` method, so this line in bk/bk's [spec_helper](https://github.com/buildkite/buildkite/blob/06550b03f31b5783003626c74805e4c8824f5cc7/spec/spec_helper.rb#L21) will need to be removed. I don't anticipate breaking any customer config with this, as we don't advertise configuring the debug mode like that, but instead nudge them to use the env variable. 
3. `.safe` method - we had this method as a catch all because the socket functionality used to throw random exceptions that we weren't expecting and weren't handling. This doesn't happen anymore, and this function isn't used anymore. 

I've tested this PR with running an rspec test and minitest test. Note that if you are testing this change locally, you'll also need to change the config in spec helper.


nprizal (approved):
Screenshot 2023-03-29 at 5 16 08 PM
 🧹✨  I didn't test it locally, but it makes sense to me the things that are being removed on this PR

PR #182, opened 2023-04-05
Release v2.1

niceking:
Bumping the version so we can release v2.1 of the gem ✨


KatieWright26 (approved):
Approving assuming the deprecation notice was signed off

PR #183, opened 2023-04-16
Gzip body of upload request

niceking:
Gzip the payload for Upload API requests, and set Content-Encoding header so the server knows how to decode the payload. Relevant changes to bk/bk to support this is in [PR #12006](https://github.com/buildkite/buildkite/pull/12006)

### DIFF
--- a/buildkite.yaml
+++ b/buildkite.yaml
@@ -5,4 +5,4 @@ steps:
       - "bundle exec rake"
     plugins:
       - docker#v3.7.0:
-          image: ruby:latest
+          image: 445615400570.dkr.ecr.us-east-1.amazonaws.com/ecr-public/docker/library/ruby:3.1.4
--- a/spec/test_collector/http_client_spec.rb
+++ b/spec/test_collector/http_client_spec.rb
@@ -46,7 +46,7 @@ RSpec.describe Buildkite::TestCollector::HTTPClient do
     allow(Net::HTTP).to receive(:new).and_return(http_double)
     allow(http_double).to receive(:use_ssl=)
 
-    allow(Net::HTTP::Post).to receive(:new).with("buildkite.localhost", {"Authorization"=>"Token token=\"my-cool-token\"", "Content-Type"=>"application/json"}).and_return(post_double)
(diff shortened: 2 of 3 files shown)


spec/test_collector/http_client_spec.rb
```diff
@@ -39,14 +39,24 @@
...
+  end
+
+  let(:compressed_body) do
+    str = StringIO.new
+
+      writer = Zlib::GzipWriter.new(str)
+      writer.write(request_body)
+      writer.close
+
+      str.string
```
yob, line 52: 👮‍♂️ rubo-james 👮‍♂️

Slightly weird indentation here

yob (commented):
I haven't tested this out, but it looks like it'll do the thing we expect. I assume we'll test it out via a Gemfile git reference in other repos? We could even load this branch directly in the Gemfile prior to merging if we'd like.

niceking:
@yob lol your review was so speedy that I hadn't even updated the other PR yet 😂 

Yup I've done an integration test of [PR #12006](https://github.com/buildkite/buildkite/pull/12006) against this change, and also sending the uncompressed payload as previously done. 

Plan is to: 
1. Merge PR #12006 and check nothing breaks
2. Merge PR #183  
3. Update Gemfile in buildkite with an updated git ref, and the branch that builds that will be sending the compressed data to a server that knows how to decompress that data

buildkite.yaml
```diff
@@ -5,4 +5,4 @@ steps:
       - "bundle exec rake"
     plugins:
       - docker#v3.7.0:
-          image: ruby:latest
+          image: 445615400570.dkr.ecr.us-east-1.amazonaws.com/ecr-public/docker/library/ruby:3.1.4
```
nprizal, line 8: Is this change intentional and what's the motivation behind this it?

niceking: One of the earlier commits on this branch failed to build due to us hitting [docker cache limits](https://buildkite.com/buildkite/rspec-buildkite-analytics/builds/647#01879745-72d5-4553-9a08-a0b4c608f699). I asked in the [serious eng campfire](https://3.basecamp.com/3453178/buckets/1933834/chats/295546306@6062773808) and James advised changing the image to use one through the pull through cache. 

I can pull this out into a separate PR if you'd prefer?

nprizal (approved):

PR #187, opened 2023-05-02
Safeguard SecureRandom.uuid

ChrisBr:
This is a follow up to https://github.com/buildkite/test-collector-ruby/issues/131.

Our test suite stubs `SecureRandom.uuid` in various places so enabling `[test-collector-ruby](https://github.com/buildkite/test-collector-ruby)` breaks our test suite.

We thought this was fixed with https://github.com/buildkite/test-collector-ruby/pull/133 but we still see this happening.

I believe the best approach is to just safe guard `SecureRandom.uuid` by storing a private reference to the method.

### DIFF
--- a/lib/buildkite/test_collector.rb
+++ b/lib/buildkite/test_collector.rb
@@ -11,7 +11,6 @@ require "net/http"
 require "time"
 require "timeout"
 require "tmpdir"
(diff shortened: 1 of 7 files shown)


blaknite:
Closed in favour of #192

PR #188, opened 2023-05-02
Fix nil pointer

ChrisBr:
I'm not entirely sure about this fix but we still get an error if we don't initialize the plugin locally.

Screenshot 2023-05-02 at 14 16 54

Potentially related to https://github.com/buildkite/test-collector-ruby/pull/125.


blaknite:
Thanks @ChrisBr! This seems like the most pragmatic fix for this bug, and gets you unblocked quickly. I'll hit go on the build and merge once it passes.

blaknite (approved):

PR #189, opened 2023-05-03
Capitalize safeguarded UUID class

blaknite:
This builds upon #187 by conforming to our preference to upcase acronyms and fixing some tests.

From the original PR:

>This is a follow up to https://github.com/buildkite/test-collector-ruby/issues/131.
>
>Our test suite stubs SecureRandom.uuid in various places so enabling [test-collector-ruby](https://github.com/buildkite/test-collector-ruby) breaks our test suite.
>
>We thought this was fixed with https://github.com/buildkite/test-collector-ruby/pull/133 but we still see this happening.
>
>I believe the best approach is to just safe guard SecureRandom.uuid by storing a private reference to the method.

### DIFF
--- a/lib/buildkite/test_collector.rb
+++ b/lib/buildkite/test_collector.rb
@@ -11,7 +11,6 @@ require "net/http"
 require "time"
 require "timeout"
 require "tmpdir"
-require "securerandom"
 
 require "active_support/core_ext/object/blank"
 require "active_support/core_ext/hash/indifferent_access"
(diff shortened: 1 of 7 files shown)


niceking (approved):
Looks good, has this been integration tested against a local bk/bk?

blaknite:
> Looks good, has this been integration tested against a local bk/bk?

Not yet!

nprizal (approved):

blaknite:
Closed in favour of #192

PR #192, opened 2023-05-17
Stop sending execution id

niceking:
Builds upon #189 and #187 

We've decided to stop generating execution IDs client side. However, that wasn't the only place the ruby test collector was using SecureRandom.uuid, so am also still including the patch that @ChrisBr submitted that @blaknite then added to 

Has been integration test against bk/bk

There'll be a follow up PR on bk/bk that also ignores any execution IDs that are sent from test collector clients but this PR can be merged and released independently of that

### DIFF
--- a/lib/buildkite/test_collector.rb
+++ b/lib/buildkite/test_collector.rb
@@ -11,7 +11,6 @@ require "net/http"
 require "time"
 require "timeout"
 require "tmpdir"
-require "securerandom"
 
 require "active_support/core_ext/object/blank"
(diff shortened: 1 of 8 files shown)


ChrisBr (approved):
Thanks!

jasmine-q:
makes sense to my not a ruby expert eyes 👀 will defer deeper diving to other rubyists

jasmine-q (approved):

gchan (approved):

Issue #194, opened 2023-05-31
Cannot debug uploader errors


dzunk:
Hello, we're trying out Buildkite right now and I'm trying to set up test reporting. Currently receiving the generic error:
```
buildkite-test_collector 2.3.0 experienced an error when sending your data, you may be missing some executions for this run.
```

It looks like this was recently introduced in https://github.com/buildkite/test-collector-ruby/pull/191, however the side effect is that the actual error message and backtrace are swallowed, so I don't have a way to see what the error is or continue debugging.

Potential solutions:
* Log error details (type, message, source location) to STDERR
* Add a `BUILDKITE_DEBUG` variable that will conditionally bypass the rescue and allow builds to fail with the full error and backtrace

If the change(s) make sense to y'all, happy to contribute one or both of them.

dzunk:
FWIW the error was:
```
VCR::Errors::UnhandledHTTPRequestError: 

================================================================================
An HTTP request has been made that VCR does not know how to handle:
  POST https://analytics-api.buildkite.com/v1/uploads
```

Which may be worth a brief mention in the docs. Need to set:
```ruby
VCR.configure do |c|
  c.ignore_hosts "analytics-api.buildkite.com"
end
```

nprizal:
Hi @dzunk thanks for raising this issue. We used to have a debugging support which we removed when we moved away from websocket https://github.com/buildkite/test-collector-ruby/pull/179. Adding a debugging back that will bypass the rescue and bubble the error is something we are reluctant to support because we don't want the collector to fail the builds. However, we can still log the error as you suggested.

As for the specific error that you got, I agree that it needs to mentioned in the doc. I will open a PR for this isue.

dzunk:
Resolved in version [2.3.1](https://github.com/buildkite/test-collector-ruby/releases/tag/v2.3.1)! Thanks for the quick action

PR #195, opened 2023-06-01
Log standard error raised during upload

nprizal:
We rescue a standard error during upload to prevent builds from failing. However, there is no way to know what the real errors are. We want to log the error while silently rescuing it so we know what went wrong.

This fixes issue https://github.com/buildkite/test-collector-ruby/issues/194

I also updated the documentation for an extra step when using VCR in a test suites, as raised by that same issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,15 @@ git commit -am "Add Buildkite Test Analytics"
 git push origin add-buildkite-test-analytics
 ```
 
+### VCR
+If your test suites use [VCR](https://github.com/vcr/vcr) to interact with HTTP, you need to allow them to make an actual request to Test Analytics.
+
+```
+VCR.configure do |c|
+  c.ignore_hosts "analytics-api.buildkite.com"
+end
+```
+
 ## 🗨️ Annotations
 
 This gem allows adding custom annotations to the span data sent to Buildkite using the [.annotate](https://github.com/buildkite/test-collector-ruby/blob/d9fe11341e4aa470e766febee38124b644572360/lib/buildkite/test_collector.rb#L64) method. For example:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -73,6 +73,15 @@ git commit -am "Add Buildkite Test Analytics"
 git push origin add-buildkite-test-analytics
 ```
 
+### VCR
+If your test suites use [VCR](https://github.com/vcr/vcr) to interact with HTTP, you need to allow them to make an actual request to Test Analytics.
```
niceking, line 77:
```suggestion
If your test suites use [VCR](https://github.com/vcr/vcr) to stub network requests, you'll need to modify the config to allow actual network requests to Test Analytics.
```

lib/buildkite/test_collector/uploader.rb
```diff
@@ -47,6 +47,7 @@ def self.upload(data)
           end
         rescue StandardError => e
           $stderr.puts "#{Buildkite::TestCollector::NAME} #{Buildkite::TestCollector::VERSION} experienced an error when sending your data, you may be missing some executions for this run."
+          $stderr.puts e
```
niceking, line 50:
```suggestion
          $stderr.puts e
          $stderr.puts "#{Buildkite::TestCollector::NAME} #{Buildkite::TestCollector::VERSION} experienced an error when sending your data, you may be missing some executions for this run."
```
Just moving the error above the log line so it is more like a stack trace

README.md
nprizal: Thanks! It's a lot more concise.

niceking (approved):

dzunk (approved):

PR #197, opened 2023-07-02
[PIE-1834] Add build_id and build_number to collector

KatieWright26:
## Description
As part of the flexi-test-grouping work, new env vars are required to create the new build views in TA. We noted that vars from other ci providers do not directly map to Buildkites, so some tweaks have been made to the github and circleci `:env`.
As these run_env variables are internal to the collector and the required ENV VARs are being passed already, there is no required update to the ruby collector docs.

Bump to version will occur in future PR.

## Context
[PIE-1834](https://linear.app/buildkite/issue/PIE-1834/patch-test-collector-ruby-to-send-new-env-attributes)

## Changes
- `build_id` has been added to this test collector
- env vars for github and circleci have been updated to ensure build_ids remain unique upon run retries.

## Deployment / Risk
Low risk

### DIFF
--- a/lib/buildkite/test_collector/ci.rb
+++ b/lib/buildkite/test_collector/ci.rb
@@ -82,8 +82,8 @@ class Buildkite::TestCollector::CI
       "url" => ENV["CIRCLE_BUILD_URL"],
       "branch" => ENV["CIRCLE_BRANCH"],
       "commit_sha" => ENV["CIRCLE_SHA1"],
-      "number" => ENV["CIRCLE_WORKFLOW_ID"],
-      "build_id" => "#{ENV["CIRCLE_WORKFLOW_WORKSPACE_ID"]}-#{ENV["CIRCLE_WORKFLOW_ID"]}"
+      "number" => ENV["CIRCLE_WORKFLOW_ID"], # There is no human readable ENV in CircleCi
+      "build_id" => ENV["CIRCLE_WORKFLOW_ID"] # Retries in CircleCi generate a new workflow id each time
     }
   end
 end
--- a/lib/buildkite/test_collector/ci.rb
+++ b/lib/buildkite/test_collector/ci.rb
@@ -16,7 +16,7 @@ class Buildkite::TestCollector::CI
   def ci_env
     return buildkite if ENV["BUILDKITE_BUILD_ID"]
     return github_actions if ENV["GITHUB_RUN_NUMBER"]
-    return circleci if ENV["CIRCLE_BUILD_NUM"]
+    return circleci if ENV["CIRCLE_WORKFLOW_ID"]
     return generic if ENV["CI"]
 
     {
@@ -82,8 +82,8 @@ class Buildkite::TestCollector::CI
       "url" => ENV["CIRCLE_BUILD_URL"],
       "branch" => ENV["CIRCLE_BRANCH"],
       "commit_sha" => ENV["CIRCLE_SHA1"],
-      "number" => ENV["CIRCLE_BUILD_NUM"],
-      "build_id" => ENV["CIRCLE_WORKFLOW_ID"]
+      "number" => ENV["CIRCLE_WORKFLOW_ID"],
+      "build_id" => "#{ENV["CIRCLE_WORKFLOW_WORKSPACE_ID"]}-#{ENV["CIRCLE_WORKFLOW_ID"]}"
     }
   end
 end
--- a/lib/buildkite/test_collector/ci.rb
+++ b/lib/buildkite/test_collector/ci.rb
@@ -38,6 +38,8 @@ class Buildkite::TestCollector::CI
       "execution_name_suffix" => ENV["BUILDKITE_ANALYTICS_EXECUTION_NAME_SUFFIX"],
       "version" => Buildkite::TestCollector::VERSION,
       "collector" => "ruby-#{Buildkite::TestCollector::NAME}",
+      "build_number" => ENV["BUILDKITE_BUILD_NUMBER"],
+      "build_id" => ENV["BUILDKITE_BUILD_ID"]
     }.compact
   end
 
@@ -58,6 +60,8 @@ class Buildkite::TestCollector::CI
       "number" => ENV["BUILDKITE_BUILD_NUMBER"],
       "job_id" => ENV["BUILDKITE_JOB_ID"],
       "message" => ENV["BUILDKITE_MESSAGE"],
+      "build_number" => ENV["BUILDKITE_BUILD_NUMBER"],
+      "build_id" => ENV["BUILDKITE_BUILD_ID"]
     }
   end
 
@@ -69,6 +73,8 @@ class Buildkite::TestCollector::CI
       "branch" => ENV["GITHUB_REF_NAME"],
       "commit_sha" => ENV["GITHUB_SHA"],
       "number" => ENV["GITHUB_RUN_NUMBER"],
+      "build_number" => ENV["GITHUB_RUN_NUMBER"],
+      "build_id" => ENV["GITHUB_RUN_ID"]
     }
   end
 
@@ -80,6 +86,8 @@ class Buildkite::TestCollector::CI
       "branch" => ENV["CIRCLE_BRANCH"],
       "commit_sha" => ENV["CIRCLE_SHA1"],
       "number" => ENV["CIRCLE_BUILD_NUM"],
+      "build_number" => ENV["CIRCLE_BUILD_NUM"],
+      "build_id" => ENV["CIRCLE_WORKFLOW_ID"]
     }
   end
 end
--- a/lib/buildkite/test_collector/ci.rb
+++ b/lib/buildkite/test_collector/ci.rb
@@ -37,8 +37,8 @@ class Buildkite::TestCollector::CI
       "execution_name_suffix" => ENV["BUILDKITE_ANALYTICS_EXECUTION_NAME_SUFFIX"],
       "version" => Buildkite::TestCollector::VERSION,
       "collector" => "ruby-#{Buildkite::TestCollector::NAME}",
-      "build_number" => ENV["BUILDKITE_ANALYTICS_NUMBER"],
-      "build_id" => ENV["BUILDKITE_BUILD_ID"]
+      "number" => ENV["BUILDKITE_ANALYTICS_NUMBER"],
+      "build_id" => ENV["BUILDKITE_ANALYTICS_BUILD_ID"]
     }.compact
   end
 
@@ -58,7 +58,7 @@ class Buildkite::TestCollector::CI
       "commit_sha" => ENV["BUILDKITE_COMMIT"],
       "job_id" => ENV["BUILDKITE_JOB_ID"],
       "message" => ENV["BUILDKITE_MESSAGE"],
-      "build_number" => ENV["BUILDKITE_BUILD_NUMBER"],
+      "number" => ENV["BUILDKITE_BUILD_NUMBER"],
       "build_id" => ENV["BUILDKITE_BUILD_ID"]
     }
   end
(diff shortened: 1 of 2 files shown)


lib/buildkite/test_collector/ci.rb
```diff
@@ -31,13 +31,14 @@ def analytics_env
...
       "commit_sha" => ENV["BUILDKITE_ANALYTICS_SHA"],
-      "number" => ENV["BUILDKITE_ANALYTICS_NUMBER"],
       "job_id" => ENV["BUILDKITE_ANALYTICS_JOB_ID"],
       "message" => ENV["BUILDKITE_ANALYTICS_MESSAGE"],
       "execution_name_prefix" => ENV["BUILDKITE_ANALYTICS_EXECUTION_NAME_PREFIX"],
       "execution_name_suffix" => ENV["BUILDKITE_ANALYTICS_EXECUTION_NAME_SUFFIX"],
       "version" => Buildkite::TestCollector::VERSION,
       "collector" => "ruby-#{Buildkite::TestCollector::NAME}",
+      "build_number" => ENV["BUILDKITE_ANALYTICS_NUMBER"],
+      "build_id" => ENV["BUILDKITE_BUILD_ID"]
```
blaknite, line 41: This one is used for overriding the default values and differs from the baked-in `BUILDKITE_BUILD_ID` it should be:
```suggestion
      "build_id" => ENV["BUILDKITE_ANALYTICS_BUILD_ID"]
```
This also makes it possible for someone using CI other than the Buildkite, GitHub Actions or CircleCI to pass in an equivalent value from their environment without overloading the meaning of the env set by Buildkite Pipelines.

lib/buildkite/test_collector/ci.rb
```diff
@@ -68,7 +70,8 @@ def github_actions
       "url" => File.join("https://github.com", ENV["GITHUB_REPOSITORY"], "actions/runs", ENV["GITHUB_RUN_ID"]),
       "branch" => ENV["GITHUB_REF_NAME"],
       "commit_sha" => ENV["GITHUB_SHA"],
-      "number" => ENV["GITHUB_RUN_NUMBER"],
+      "build_number" => ENV["GITHUB_RUN_NUMBER"],
+      "build_id" => ENV["GITHUB_RUN_ID"]
```
blaknite, line 74: According to the [GitHub docs](https://docs.github.com/en/actions/learn-github-actions/variables), `GITHUB_RUN_ID` doesn't change upon re-run. I'm not sure this is suitable for our purposes. We can combine `GITHUB_RUN_ATTEMPT` with `GITHUB_RUN_ID` to get a unique value upon re-run.

We could also used the value created on line 69. It uses different env but achieves the same result.

lib/buildkite/test_collector/ci.rb
```diff
@@ -79,7 +82,8 @@ def circleci
       "url" => ENV["CIRCLE_BUILD_URL"],
       "branch" => ENV["CIRCLE_BRANCH"],
       "commit_sha" => ENV["CIRCLE_SHA1"],
-      "number" => ENV["CIRCLE_BUILD_NUM"],
+      "build_number" => ENV["CIRCLE_BUILD_NUM"],
+      "build_id" => ENV["CIRCLE_WORKFLOW_ID"]
```
blaknite, line 86: I think I may have got my understanding of workflows wrong when I set the default value for `key` on circle CI.

I think you're correct in choosing `CIRCLE_WORKFLOW_ID` as synonymous with `BUILDKITE_BUILD_ID`. We may need to fix line 81 to jus use `CIRCLE_WORKFLOW_ID` 🤔 What do you think?

If I'm understanding the relationship of circle workflows and builds correctly, I think `CIRCLE_BUILD_NUMBER` is closer to `BUILDKITE_JOB_NUMBER` (if it existed) than it is to `BUILDKITE_BUILD_NUMBER`.

Maybe we can chat about this one in breakout?

blaknite (commented):
Thanks @KatieWright26! I've suggested a couple of changes and I think there's some nuance in these values that could use a little discussion in breakout.

KatieWright26:
> Thanks @KatieWright26! I've suggested a couple of changes and I think there's some nuance in these values that could use a little discussion in breakout.

Yeah I've been thinking about this ticket a bunch too and have some stuff to talk about, I think breakout sounds good.

lib/buildkite/test_collector/ci.rb
KatieWright26: Ok I've had a hoon around the circleci docs, and also asked ChatGPT and it looks like circleci doesn't have an equivelent of `BUILDKITE_BUILD_NUMBER` or `GITHUB_RUN_ATTEMPT`. ChatGPT says you could use `CIRCLE_BUILD_NUM` with custom logic to work it out, but I think this might be lies given it gave me a totally wrong answer the first time I asked.
I *think* CircleCI creates a whole new workflow id on retry given this [answer on the community pages](https://discuss.circleci.com/t/detect-rerun-of-a-job/44735), so `WORKFLOW_ID` might be good enough?

lib/buildkite/test_collector/ci.rb
```diff
@@ -31,13 +31,14 @@ def analytics_env
       "url" => ENV["BUILDKITE_ANALYTICS_URL"],
```
KatieWright26, line 31: @blaknite is this `:key` going to work for our needs? Is this a static value?

lib/buildkite/test_collector/ci.rb
KatieWright26: I've implemented this, but I can't find `BUILDKITE_ANALYTICS_BUILD_ID` anywhere in the docs, is this a new variable we need to add?

lib/buildkite/test_collector/ci.rb
KatieWright26: I checked with an old colleague who still uses circleci and she did some digging around and pretty much validated the community pages answer. `CIRCLE_WORKFLOW_WORKSPACE_ID` does not change on rerun, but `CIRCLE_WORKFLOW_ID` does, so we can use these in tandem to create the equivalent of a BUILD_ID.

example data:
For the 0th parallel run in the **RSpec** workflow for that CI run, those are:
```
  CIRCLE_WORKFLOW_ID=aaaa... <--- match
  CIRCLE_WORKFLOW_JOB_ID=bbbb...
  CIRCLE_WORKFLOW_WORKSPACE_ID=aaaa... <--- match
```
For the 1st parallel run in the **RSPec** workflow for that CI run, those are the same:
```
  CIRCLE_WORKFLOW_ID=aaaa...
  CIRCLE_WORKFLOW_JOB_ID=bbbb...
  CIRCLE_WORKFLOW_WORKSPACE_ID=aaaa...
```
For the **Rubocop** workflow for that CI run, those are:
```
  CIRCLE_WORKFLOW_ID=aaaa...
  CIRCLE_WORKFLOW_JOB_ID=cccc...
  CIRCLE_WORKFLOW_WORKSPACE_ID=aaaa...
```
If you **re-run** CI and look at the **Rubocop** workflow, they’re:
```
  CIRCLE_WORKFLOW_ID=eeee...
  CIRCLE_WORKFLOW_JOB_ID=dddd...
  CIRCLE_WORKFLOW_WORKSPACE_ID=aaaa...
```

nprizal (commented):
> ## Changes
> - build_id and build_number have been added to this test collector
> - number has been renamed to build_number

Looking at the latest changes, it seems like you are keeping the `number` and changed the value for `circleci`. Can you update the PR description?

lib/buildkite/test_collector/ci.rb
blaknite: Is `CIRCLE_WORKFLOW_ID` unique across all the things or only within a workspace? ie can we use it on its own?

KatieWright26: Argh I see I got build_number and build_id confused here; yes `CIRCLE_WORKFLOW_ID` is unique, but circleCi doesn't have a concept of an incrementer variable like buildkite (BUILDKITE_BUILD_NUMBER) and github does (GITHUB_RUN_NUMBER). 
On rerun, the `CIRCLE_WORKFLOW_ID` changes, so without the CIRCLE_WORKFLOW_WORKSPACE_ID you can't tell the difference between a new workflow and a rerun.

Do we care about this? Do we set build_number and build_id to the same value here?

lib/buildkite/test_collector/ci.rb
```diff
@@ -69,17 +71,19 @@ def github_actions
...
   def circleci
     {
       "CI" => "circleci",
-      "key" => "#{ENV["CIRCLE_WORKFLOW_ID"]}-#{ENV["CIRCLE_BUILD_NUM"]}",
+      "key" => ENV["CIRCLE_WORKFLOW_ID"],
       "url" => ENV["CIRCLE_BUILD_URL"],
       "branch" => ENV["CIRCLE_BRANCH"],
       "commit_sha" => ENV["CIRCLE_SHA1"],
-      "number" => ENV["CIRCLE_BUILD_NUM"],
+      "number" => ENV["CIRCLE_WORKFLOW_ID"], # There is no human readable ENV in CircleCi
```
meghan-kradolfer, line 85: I see that `CIRCLE_WORKFLOW_ID` is a string, would this create any issues with the cast we are doing in the query for `find_build_ids`? Could this string be a UUID rather than just a count that happens in BK?

meghan-kradolfer: Might also mean that the `#` we are appending to the number in the UI might look a bit weird when it is something like `Build #550e8400-e29b-41d4-a716-446655440000`

KatieWright26: I'm less worried about the `#` look, as we're constrained by circleci's vars on offer. We could always truncate the build number or something. 

I did double check the cast and you're correct that it breaks, as we can't cast the string uuid to an integer. Having a look for ways around this atm; @blaknite probably has ideas.

KatieWright26: ChatGPT and I got this working with the ammendment to the find_build_id SQL:
```
SELECT
            env->>'build_id' AS build_id,
            (CASE
              WHEN env->>'number' ~ '^-?\d+$' THEN CAST(env->>'number' AS INTEGER)
              WHEN env->>'number' ~ '^-?\d+(\.\d+)?$' THEN CAST(env->>'number' AS NUMERIC)
              ELSE NULL
            END) AS number,
            created_at
          FROM runs
```
Screenshot 2023-07-05 at 11 07 44 (2)

lib/buildkite/test_collector/ci.rb
blaknite: My rudimentary understanding is that CircleCI has "workflows" and "workflow instances" which are synonymous with "pipelines" and "builds". If my understanding is correct, `CIRCLE_WORKFLOW_ID` is exactly the variable we're after.

Does my understanding match yours?

lib/buildkite/test_collector/ci.rb
```diff
@@ -69,17 +71,19 @@ def github_actions
       "branch" => ENV["GITHUB_REF_NAME"],
       "commit_sha" => ENV["GITHUB_SHA"],
       "number" => ENV["GITHUB_RUN_NUMBER"],
+      "build_id" => "#{ENV["GITHUB_RUN_ID"]}-#{ENV["GITHUB_RUN_ATTEMPT"]}"
```
blaknite, line 74: I think we might also need `GITHUB_ACTION` added to this value.

From the [GitHub docs](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables):
> A unique number for each workflow run **within** a repository.
That sounds like it's not unique outside the repo. There's nothing preventing someone from using TA with multiple repos so it might not be unique enough for our purposes.

What do you think?

KatieWright26: HMM GITHUB_ACTION doesn't sound unique either though:

> The name of the action currently running, or the id of a step. For example, for an action, __repo-owner_name-of-action-repo...
> GitHub removes special characters, and uses the name __run when the current step runs a script without an id...
> If you use the same script or action more than once in the same job, the name will include a suffix that consists of the sequence number preceded by an underscore [e.g] `__run_2`

I reckon using the GITHUB_REPOSITORY_ID would be the safest way to guarantee uniqueness given run_ids are scoped to them.

KatieWright26:
Closing this as no longer applicable to flexi-test grouping after Q3 pivot

PR #198, opened 2023-08-28
Add support for Codeship

swebb:
Environment variables documented [here](https://docs.cloudbees.com/docs/cloudbees-codeship/latest/pro-builds-and-configuration/environment-variables#_default_environment_variables).

### DIFF
--- a/lib/buildkite/test_collector/ci.rb
+++ b/lib/buildkite/test_collector/ci.rb
@@ -17,6 +17,7 @@ class Buildkite::TestCollector::CI
     return buildkite if ENV["BUILDKITE_BUILD_ID"]
     return github_actions if ENV["GITHUB_RUN_NUMBER"]
     return circleci if ENV["CIRCLE_BUILD_NUM"]
+    return codeship if ENV["CI_NAME"] == "codeship"
     return generic if ENV["CI"]
 
     {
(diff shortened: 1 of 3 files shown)


lib/buildkite/test_collector/ci.rb
```diff
@@ -82,4 +83,16 @@ def circleci
...
   end
+
+  def codeship
+    {
+      "CI" => "codeship",
+      "key" => "#{ENV["CI_BUILD_ID"]}",
+      "url" => ENV["CI_PULL_REQUEST"],
+      "branch" => ENV["CI_BRANCH"],
+      "commit_sha" => ENV["CI_COMMIT_ID"],
+      "number" => nil,
```
KatieWright26, line 94: I had a look and it seems there's no equivalent to BUILD_NUMBER. Is it ok to have this blank? What about a `CI_TIMESTAMP`?

lib/buildkite/test_collector/ci.rb
```diff
@@ -82,4 +83,16 @@ def circleci
       "number" => ENV["CIRCLE_BUILD_NUM"],
     }
   end
+
+  def codeship
+    {
+      "CI" => "codeship",
+      "key" => "#{ENV["CI_BUILD_ID"]}",
```
gchan, line 90: Is the ID globally unique or only unique to the project?

Do we want to incorporate the `CI_PROJECT_ID`?

gchan: I had a peek at their API, it seems projects and builds are globally unique with UUID-shaped identifiers. 👍️

lib/buildkite/test_collector/ci.rb
swebb: TBH I'm not sure what that controls. According to the [docs](https://buildkite.com/docs/test-analytics/ci-environments#recommended-environment-variables) it's not a "recommended" environment variable. 🤷🏻

swebb:
[Here's](https://buildkite.com/organizations/buildkite/analytics/suites/packagecloud/runs/b47c2013-6e74-4329-b4a9-fc8c144d7e03) and example run.

gchan (approved):
LGTM 🚀

PR #200, opened 2023-08-31
Test gem under every supported Ruby version

gchan:
Update build pipeline to run tests under multiple Ruby versions

All minor Ruby versions from 3.2 to 2.3 will be tested.

The gemspec says Ruby versions from 2.3.x are supported, however
some modern Ruby syntax was introduced which breaks compatibility
for older Ruby versions.

### DIFF
--- a/buildkite.yaml
+++ b/buildkite.yaml
@@ -7,6 +7,7 @@ steps:
       - docker#v3.7.0:
           image: "445615400570.dkr.ecr.us-east-1.amazonaws.com/ecr-public/docker/library/ruby:{{matrix}}"
     matrix:
+      - "latest"
       - "3.2.2"
       - "3.1.4"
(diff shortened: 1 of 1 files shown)


buildkite.yaml
```diff
@@ -1,8 +1,31 @@
...
+      - "3.0.6"
+      - "2.7.8"
+  - group: ":rspec: Legacy Ruby :ruby:"
+    steps:
+      - label: ":rspec: Tests :ruby: {{matrix}}"
+        command:
+          - "gem install bundler:2.3.25"
+          - "bundle"
+          - "bundle exec rake"
+        soft_fail: true
```
gchan, line 21: Temporarily allow these builds to fail. Remove when we restore comparability.

KatieWright26 (approved):

Issue #199, opened 2023-08-30
Gem isn't ruby 2.3 compatible


alagos:
That's what is defined [in the gemspec](https://github.com/buildkite/test-collector-ruby/blob/main/buildkite-test_collector.gemspec#L25), but it has ruby 2.6 syntax [like this](https://github.com/buildkite/test-collector-ruby/blob/main/lib/buildkite/test_collector/minitest_plugin/trace.rb#L77) which makes the gem to fail in older versions (ruby 2.4 in my case), so it should be updated the gemspec or change that syntax to make it compatible with older ruby versions.

gchan:
Hi @alagos , it sounds like your preference is for Ruby 2.4 to remain supported (let me know otherwise!).

I'll check with the team to see whether [EOL Ruby versions](https://www.ruby-lang.org/en/downloads/branches/) should continue to be supported.

alagos:
Well, if ruby 2.4 can be supported, that would be awesome. We're still in the process to upgrade to supported ruby versions, but in the meantime would be nice to have access to this gem.

gchan:
Hi @alagos , I'm working on restoring comparability for Ruby 2.3 over at #201 , let me know if you have any feedback or comments!

Support for Ruby 2.4 will not be permanent and we will likely have a plan to gradually drop support for older versions of Ruby. 

How long do you think your team requires Ruby 2.4 support?

alagos:
@gchan wow, that's great. 
We're in the process to upgrade our stack (a year ago we were with ruby 2.2), so hopefully a year from now.

gchan:
Hi @alagos , we've published a new version of the gem which supports Ruby versions 2.3 and above.

https://github.com/buildkite/test-collector-ruby/releases/tag/v2.3.2

Please note support for EOL'ed versions of Ruby may be removed in the future.

alagos:
Sure, I'll keep in mind.
Thanks for all

PR #201, opened 2023-08-31
Restore compatibility with older Ruby versions

gchan:
The gemspec file states Ruby versions 2.3 and above is supported but there is code which breaks compatibility with versions 2.3, 2.4 and 2.5.

See https://github.com/buildkite/test-collector-ruby/issues/199

This PR removes the use Ruby features/syntax that were introduced in 2.4, 2.5 and 2.6.

### DIFF
--- a/lib/buildkite/test_collector/minitest_plugin/trace.rb
+++ b/lib/buildkite/test_collector/minitest_plugin/trace.rb
@@ -74,7 +74,7 @@ module Buildkite::TestCollector::MinitestPlugin
         # remove the first line of message from the first failure
         # to avoid duplicate line in Test Analytics UI
         messages = strip_invalid_utf8_chars(failure.message).split("\n")
-        messages = messages[1..] if index.zero?
+        messages = messages[1..-1] if index.zero?
 
         {
           expanded: messages,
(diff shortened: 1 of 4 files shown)


rafaelmagu (approved):

KatieWright26 (approved):
1 question but LGTM

lib/buildkite/test_collector/minitest_plugin/trace.rb
```diff
@@ -74,7 +74,7 @@ def failure_expanded
         # remove the first line of message from the first failure
         # to avoid duplicate line in Test Analytics UI
         messages = strip_invalid_utf8_chars(failure.message).split("\n")
-        messages = messages[1..] if index.zero?
+        messages = messages[1..-1] if index.zero?
```
KatieWright26, line 77: Is this change just for clarity?

gchan: Functionally the syntax is equivalent but Ruby endless ranges was only introduced in Ruby 2.6. 

https://rubyreferences.github.io/rubychanges/2.6.html#endless-range-1

PR #207, opened 2023-10-15
[PIE-2044] Output a summary of TA links in rspec log

meghan-kradolfer:
### Description

Output test analytics links within the summary of an rspec run. Using the rspec `dump_failures` method we can output a list of all the failed examples within a run. 

Links will only output links if:
- more than 1 test has failed
- a test analytics token has been provided
- a suite url has been received from the upload

The test link url is created by getting the `suite_url` from the [from the metadata endpoint within BK](https://github.com/buildkite/buildkite/pull/14288/files#diff-f1b550c16dfbb29caae2f9eebd8c9fbda11129ab77dde799a2097175a91b700dR91), the suite_url is currently only populated when the feature flag `analytics_job_log_test_links` is turned on, when the flag is off the `dump_failures` will return empty and not output will appear. The `suite_url` is then appended with the `scope_name_digest` [as the test identifier](https://github.com/buildkite/buildkite/pull/14287)

https://github.com/buildkite/test-collector-ruby/assets/25017977/b757cf16-daf2-48c7-9a30-f4dd5c91fe18

### Context 
[PIE-2044](https://linear.app/buildkite/issue/PIE-2044/output-a-summary-of-ta-links-in-rspec-log)

### Verification

Integration testing using buildkite/buildkite

Renders failures correctly:
- [x] Feature flag is on (metadata success, suite_url present)

Does not render correctly:
- [x] Feature flag is off  (metadata success, suite_url nil)

Fails gracefully:
- [x] Feature flag is on and metadata endpoint throws an error or returns a non 200 code
- [x] Feature flag is off  and metadata endpoint throws an error or returns a non 200 code


lib/buildkite/test_collector/uploader.rb
```diff
@@ -40,7 +44,7 @@ def self.upload(data)
       Thread.new do
         response = begin
           upload_attempts ||= 0
-          http.post_json(data)
+          @response = http.post_json(data)
```
meghan-kradolfer, line 47: We can add the suite_url to the response from the `post_json` which comes from `Analytics::API::UploadsController.create` within BK.

I'm not 100% on this part, and how I am assigning the response here.. keen to get feedback on it

meghan-kradolfer: Draft PR for the BK part of this here: https://github.com/buildkite/buildkite/pull/14288

lib/buildkite/test_collector/test_links_plugin/trace.rb
nprizal, line 1: Looks like this module is not used other than for test, can we remove this module? I can see that `TestLinks::Reporter` generates the digest from attributes in the result object (`notification.example`), I don't think we need to trace that result at all.

meghan-kradolfer: Ah yup! I think I was using it and forgot to remove. Is it better to be using the trace? Or ok to be generating it in here?

nprizal: I don't think we need to use the trace for this one, the `TestLinks::Reporter` is still small enough to be generating the digest itself.

lib/buildkite/test_collector/test_links_plugin/reporter.rb
```diff
@@ -0,0 +1,56 @@
...
+    def example_failed(notification)
+      example = notification.example
+
+      @failed_examples << {
+        scope: example.example_group.metadata[:full_description],
+        name: example.description
+      }
+    end
+
+    def dump_summary(_notification)
```
niceking, line 22: I think you could hook into `dump_failures` instead of `dump_summary` and then avoid some of the checks you're currently doing? 

https://rspec.info/documentation/3.7/rspec-core/RSpec/Core/Formatters/Protocol.html#dump_failures-instance_method

niceking (commented):
i've got some other thoughts on this but i think it might be better to do in-sync rather than async?

lib/buildkite/test_collector/test_links_plugin/reporter.rb
```diff
@@ -0,0 +1,52 @@
...
+
+module Buildkite::TestCollector::TestLinksPlugin
+  class Reporter
+    RSpec::Core::Formatters.register self, :dump_failures
+
+    def initialize(output)
+      @output = output
+    end
+
+    def dump_failures(notification)
```
meghan-kradolfer, line 11: @niceking I've updated this to `dump_failures` now! I still need to do the check on line 13 though as this method is called even when there are no failures 🤷 But it does cut down a lot of the extra stuff I was doing with assigning the failures to an array.

lib/buildkite/test_collector/uploader.rb
```diff
@@ -51,5 +51,14 @@ def self.upload(data)
         end
       end
     end
+
+    def self.summary
+      return false unless Buildkite::TestCollector.api_token
+
+      http = Buildkite::TestCollector::HTTPClient.new("#{Buildkite::TestCollector.url}/summary")
```
meghan-kradolfer, line 58: @niceking this has been updated here: [PIE-2047](https://github.com/buildkite/buildkite/pull/14288/files#diff-f1b550c16dfbb29caae2f9eebd8c9fbda11129ab77dde799a2097175a91b700dR91 ). I'm using the url `.../uploads/summary` as we send the uploads url through to the test collector, so this means if we created a new url for summary we will also need to send this through, maybe I am just being lazy here though and it is better to send through another unique url?

lib/buildkite/test_collector/http_client.rb
```diff
@@ -41,6 +41,20 @@ def post_json(data)
       http.request(contact)
     end
 
+    def summary_links
```
meghan-kradolfer, line 44: I'm not sure if summary / summary_links is the correct name here. I was imaging in the future we could send through some extra links or into in this to output in the log

lib/buildkite/test_collector/test_links_plugin/reporter.rb
```diff
@@ -0,0 +1,52 @@
...
+      # Do not display summary if no failed examples
+      return false unless notification.failed_examples.length.positive?
+
+      # Check if a Test Analytics token is set
+      return false unless Buildkite::TestCollector.api_token
+
+      # If the suite_url fails to be generated, then we are unable to create the test links
+      return false unless (url = suite_url)
```
niceking, line 19: `unless` is used to evaluate a boolean condition, however you're doing an assignment here, so this condition will always evaluate to true. 

What are you intending to check here? The `suite_url` method doesn't return a boolean. I think you're trying to check if a `suite_url` was able to be fetched - but the way this method works currently doesn't allow you to do that? I've also got a comment later on about this method, I think in general it needs so redesigning, let me know if you wanna pair!

lib/buildkite/test_collector/test_links_plugin/reporter.rb
```diff
@@ -0,0 +1,52 @@
...
+  class Reporter
+    RSpec::Core::Formatters.register self, :dump_failures
+
+    def initialize(output)
+      @output = output
+    end
+
+    def dump_failures(notification)
+      # Do not display summary if no failed examples
+      return false unless notification.failed_examples.length.positive?
```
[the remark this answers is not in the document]
meghan-kradolfer: Yup failed_examples is an Array! So we would have to do `return unless !notification.failed_examples.empty?`, but is this considered a 'double negative', would it be better to do something like `return unless notification.failed_examples.present?`

lib/buildkite/test_collector/test_links_plugin/reporter.rb
meghan-kradolfer: The `suite_url` on line 35 will return `nil` if the feature flag is off, so this is checking for that. It is being assigned to `url` and if `url` is nil it will return false. Maybe I am just missing the check here for if `suite_url` was unable to be fetched?

meghan-kradolfer: Also my comment above this line is a bit misleading!

lib/buildkite/test_collector/test_links_plugin/reporter.rb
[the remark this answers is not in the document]
niceking: yeah or `return if notification.failed_examples.empty?`

lib/buildkite/test_collector/test_links_plugin/reporter.rb
niceking: Yeah I think this bit is a bit confusing! It might make more sense once you rearchitect how the suite url is fetched? Give me a bell if you wanna pair :)

lib/buildkite/test_collector/test_links_plugin/reporter.rb
```diff
@@ -0,0 +1,65 @@
...
+      return unless Buildkite::TestCollector.api_token
+
+      metadata = fetch_metadata
+      
+      return if metadata.nil?
+
+      url = metadata.fetch('suite_url')
+
+      # If the suite_url does not exist, then we are unable to create the test links
+      return unless url
```
niceking, line 25: Ok I had a bit of a think about this and I think this section should be:

```
metadata = fetch_metadata

# return if metadata was not fetched successfully
return if metadata.nil?

# return if metdata does not contain suite url
return if metadata["suite_url"].nil?
```

And then line 30 can become 
```
failed_example_output(example, metadata["suite_url"])
```
as there's no need to assign the suite_url to the url field anymore because it's not doing an outbound request (you can do the assignment if you wish but there isn't a functional reason to avoid it)

and lines 56-58 can go back to 
```
JSON.parse(response.body) if response.code == 200 
```

it would be worth writing tests for all scenarios of:
1. metadata wasn't requested
2. metadata was requested but there was an error thrown in the process
3. metadata was requested but server responds with non 200 response
4. metadata was requested but suite_url returned is nil
5. metadata was requested and suite_url is present

it would probably also be worth integration testing all these scenarios - give me a bell if you need a hand with that

meghan-kradolfer: Perfect! I will give this ago. Also, I realised that it the response code is a string not boolean, that was why our original test failed

meghan-kradolfer:
@niceking Those updates have been made and new specs added, would you be able to have another look please?

spec/test_collector/test_links/reporter_spec.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+require 'buildkite/test_collector/test_links_plugin/reporter'
+
+RSpec.describe Buildkite::TestCollector::TestLinksPlugin::Reporter do
+  let(:examples) { RSpec.world.filtered_examples }
```
niceking, line 6: I've never seen this `RSpec.world` before... is it this thing? https://rspec.info/documentation/3.9/rspec-core/RSpec/Core/World.html If so, perhaps refactor to remove? 

I also don't know if you've seen this, but there are some helper methods to mock example objects in this file: https://github.com/buildkite/test-collector-ruby/blob/c57c586b03de84da7faf725e701ef8d55e2844db/spec/support/rspec_example_trace_helpers.rb

spec/test_collector/test_links/reporter_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+  let(:http_client) { double('Buildkite::TestCollector::HTTPClient') }
+  let(:suite_url) { 'https://example.com/suite/12345' }
+  let(:response) { { suite_url: suite_url }.to_json }
+  let(:io) { StringIO.new }
+  let(:reporter) { Buildkite::TestCollector::TestLinksPlugin::Reporter.new(io) }
+
+  before do
+    allow(Buildkite::TestCollector::HTTPClient).to receive(:new).and_return(http_client)
+    Buildkite::TestCollector.configure(
```
niceking, line 15: 💅 just a small style thing - would be good to have a line break here just for clarity

spec/test_collector/test_links/reporter_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      url: 'http://fake.buildkite.localhost/v1/uploads'
+    )
+  end
+
+  after do
+    reset_io(io)
+  end
+
+  context 'when tests have failed' do
+    let!(:failed_example) do
```
niceking, line 27: I think you could create a failed example using that rspec example trace helper class I mentioned earlier, or perhaps extending it to meet your needs for this test?

spec/test_collector/test_links/reporter_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      OpenStruct.new(
+        example_group: OpenStruct.new(metadata: { full_description: 'I love to eat pies' }),
+        description: 'mince and cheese',
+        execution_result: OpenStruct.new(status: :failed)
+      )
+    end
+    let!(:notification) { RSpec::Core::Notifications::SummaryNotification.new(10.0, examples, [failed_example]) }
```
niceking, line 34: I think it would be good to double this `notification` object similarly to how the `example` object was doubled

spec/test_collector/test_links/reporter_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      it 'does not request metadata' do
+        reporter.dump_failures(notification)
+        expect(Buildkite::TestCollector::HTTPClient).not_to receive(:new)
+      end
+    end
+
+    context 'fetch_metadata does not return a suite_url' do
+      it 'does not render summary' do
+        allow(http_client).to receive(:metadata).and_return(OpenStruct.new(code: '200',
```
niceking, line 54: again just a 💅 thing: consider formatting like 

```
allow(http_client).to receive(:metadata)
                  .and_return(OpenStruct.new(code: '200', body: { suite_url: nil }.to_json))
```

(ok i cant get it to work super nicely in this view, but bascially to indent the `.to receive` and the `.and_return` at the same level)

niceking (commented):
Nice!! The test cases look great, and great coverage of all things ✨ Just a few comments about test set up and setting up mock objects but pretty much there!!!

How did you go with integration testing btw?

spec/test_collector/test_links/reporter_spec.rb
meghan-kradolfer: Ah yup, I missed that warning! I can pretty much do the same thing using the helper method so have removed

meghan-kradolfer:
@niceking updates to those tests have been made. I also updated the description and added some checks for the integration testing I've done

lib/buildkite/test_collector/test_links_plugin/reporter.rb
```diff
@@ -0,0 +1,60 @@
...
+        failed_example_output(example, metadata['suite_url'])
+      end.join("\n")
+
+      @output << "\n\n"
+    end
+
+    private
+
+    def generate_scope_name_digest(scope, name)
+      Digest::SHA256.hexdigest(scope.to_s + name.to_s)
```
niceking, line 38: I'm a bit surprised that `scope` and `name` here aren't strings already? They seem like they should be from lines 42-32?

niceking (approved):
Nice!!! This looks great!!! Have left one comment on a bit that surprised me but its neither here nor there so approving 😀 

Can't wait to see this in action! Let me know if you need a hand releasing this gem or bundling bk/bk with the new version 😁

lib/buildkite/test_collector/test_links_plugin/reporter.rb
meghan-kradolfer: Do we know for sure these will be strings? I see that we are converting to strings here https://github.com/buildkite/buildkite/blob/main/app/models/analytics/execution/batch_creator.rb#L165

So maybe there is a chance that it could contain numbers?

niceking: Interesting...I reckon that it's done that way in the execution batch creator to handle nil cases? Because test collector ruby always sends values for scope and name, but other test collectors / users won't necessarily

niceking: I had a bit of a dig through the rspec docs and it does seem to indicate they should be strings: https://rubydoc.info/gems/rspec-core/RSpec/Core/Example 

Anyways it's not a huge deal - happy for you to merge as is, I was just a little surprised to see those `to_s` calls there!

meghan-kradolfer: Ohh interesting! That makes sense about them coming from other collectors. I will do a few tests with some random non-string test names and if its all good I will remove!

meghan-kradolfer: Tested with numbers and it converts to string, will remove the `to_s`

PR #208, opened 2023-10-29
Bump version to v2.3.3

meghan-kradolfer:
What's changed:
- [Output a summary of TA links in rspec log](https://github.com/buildkite/test-collector-ruby/pull/207)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v2.3.3
 
-- Added feature to output links to Test Analytic tests within the log
+- Added feature to output links to Test Analytic tests within the log @meghan-kradolfer
 
 ## v2.3.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v2.3.3
+
+- Added feature to output links to Test Analytic tests within the log
+
 ## v2.3.2
 
 - Add support for Codeship #198 - @swebb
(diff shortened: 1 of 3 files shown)


niceking:
Hey! This looks correct, but the Gemfile.lock is bumping up the version number of a lot of gems that aren't test collector ruby. Because we'll be including this gem in bk/bk, it'll then bump up the same gems over in that repository too, which might not be safe to do in production without some thorough testing, especially in core libraries like activesupport.

Could you please regenerate the Gemfile.lock with a smaller changeset? You should be able to check out the Gemfile.lock from main and then run `bundle update buildkite-test_collector --conservative` which will only update dependant libraries if they absolutely have to be updated

niceking:
Also, we've changed to have bk/bk bundle test collector ruby directly from the github source, rather than a gem version, so you also don't *have* to do release, you could just update the gem in bk/bk and check that the commit sha in the Gemfile.lock points to newest commit in this repo

meghan-kradolfer:
> Could you please regenerate the Gemfile.lock with a smaller changeset? You should be able to check out the Gemfile.lock from main and then run `bundle update buildkite-test_collector --conservative` which will only update dependant libraries if they absolutely have to be updated

Done!

niceking (approved):
✨ 💖

nprizal (commented):
I understand that we don't have versioning rules at the moment, but maybe we should start follow [Semantic Versioning](https://semver.org)? Since this is a feature, I'd suggest bumping the feature to `2.4.0`

meghan-kradolfer:
I'm going to close this PR and make a new one to update to `2.4.0`

PR #209, opened 2023-11-06
Add signifier to link outputs

meghan-kradolfer:
The test links dump is a bit hidden at the bottom of the rspec output. This means it could be easily missed by users scrolling through the output. This PR jazz's the links up to draw more attention by adding some emoji's an text colours. I also renamed the formatter as the previous naming was a bit misleading.

The new links display as below within Buildkite's job logs:

Screenshot 2023-11-08 at 9 29 39 AM


lib/buildkite/test_collector/test_links_plugin/formatter.rb
```diff
@@ -43,7 +45,7 @@ def failed_example_output(example, url)
       name = example.description
       scope_name_digest = generate_scope_name_digest(scope, name)
       test_url = "#{url}/tests/#{scope_name_digest}"
-      "\x1b[31m#{%(\x1b]1339;url=#{test_url};content="#{scope} #{name}"\x07)}\x1b[0m"
+      "🔗 \x1b[37m#{%(\x1b]1339;url=#{test_url};content="#{scope} #{name}"\x07)}\x1b[m"
```
KatieWright26, line 48: Was it possible in the end to underline the links? I'd really love if we can implement that convention now rather than later

meghan-kradolfer: Yup, we can force an underline with ANSI so the updated UI looks like this:
Screenshot 2023-11-07 at 9 21 29 AM

KatieWright26 (approved):
Thanks for making that change!

lib/buildkite/test_collector/test_links_plugin/formatter.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Buildkite::TestCollector::TestLinksPlugin
-  class Reporter
+  class Formatter
```
niceking, line 4: oh yes great name change ✨

lib/buildkite/test_collector/test_links_plugin/formatter.rb
```diff
@@ -23,7 +23,9 @@ def dump_failures(notification)
       # return if suite url is nil
       return if metadata['suite_url'].nil?
 
-      @output << "\n\nTest Analytics failures:\n\n"
+      @output << "\n\n🔥 \x1b[31mTest Analytics failures:\n"
```
niceking, line 26: hahahaha do we need MORE fire???

niceking (approved):
Looks great!

lib/buildkite/test_collector/test_links_plugin/formatter.rb
meghan-kradolfer: What are you imagining?? Open to any suggestions!

jasmine-q: I personally like a 🔥 Test Analytics failures 🔥 
or a ⚠️ Test Analytics failures ⚠️

meghan-kradolfer: I like the double flame!
Screenshot 2023-11-08 at 9 28 00 AM

PR #213, opened 2024-02-13
Add option to only send failures

francoiscampbell:
In our use-cases, we're not interested in looking at successful test cases (and being billed for them), so we'd like an option to only send failures to the API.

This PR also fixes the `fake_example` helper that was ignoring the `status` argument and always creating failing examples.

I've tested this in our org and confirmed that it indeed only sends failed tests.

### DIFF
--- a/lib/buildkite/test_collector.rb
+++ b/lib/buildkite/test_collector.rb
@@ -39,15 +39,17 @@ module Buildkite
       attr_accessor :session
       attr_accessor :tracing_enabled
       attr_accessor :artifact_path
+      attr_accessor :failures_only
       attr_accessor :env
       attr_accessor :batch_size
     end
 
-    def self.configure(hook:, token: nil, url: nil, tracing_enabled: true, artifact_path: nil, env: {})
+    def self.configure(hook:, token: nil, url: nil, tracing_enabled: true, artifact_path: nil, failures_only: false, env: {})
       self.api_token = (token || ENV["BUILDKITE_ANALYTICS_TOKEN"])&.strip
       self.url = url || DEFAULT_URL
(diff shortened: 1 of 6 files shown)


lib/buildkite/test_collector.rb
```diff
@@ -39,15 +39,17 @@ class << self
       attr_accessor :session
       attr_accessor :tracing_enabled
       attr_accessor :artifact_path
+      attr_accessor :failures_only
       attr_accessor :env
       attr_accessor :batch_size
     end
 
-    def self.configure(hook:, token: nil, url: nil, tracing_enabled: true, artifact_path: nil, env: {})
+    def self.configure(hook:, token: nil, url: nil, tracing_enabled: true, artifact_path: nil, failures_only: false, env: {})
```
francoiscampbell, line 47: I don't know if adding the arg in the middle of the sequence will cause failures, but the readme says to use named args, so it should be fine?

pda: That should be fine, I don't believe there's any way any caller could be using this method in an order-dependent way 👍🏼

pda:
@francoiscampbell Out of product-and-pricing curiosity, is the “and being billed for them” aspect the only significant downside to uploading all tests including those that passed? Or are there other reasons too?

There's a lot of value that we can today and will in future offer that relies on full awareness of the test suite (for example, flaky test detection needs to see that a test passed and failed without code changes in between). We'd love to get more feedback on how pricing fits into that. Perhaps we can can get in touch privately to get your thoughts?

francoiscampbell:
After discussing with the Buildkite Test Analytics team, this option isn't compatible with the roadmap for the product.

PR #215, opened 2024-03-06
Trace spans can be filtered by a minimum duration

pda:
Introduce the ability to filter trace spans by a minimum duration, so that fast spans can be discarded and only slow (interesting) spans are kept in memory and uploaded.

Configurable using `BUILDKITE_ANALYTICS_TRACE_MIN_SECONDS` (decimal value) in environment.

### DIFF
--- a/lib/buildkite/test_collector/tracer.rb
+++ b/lib/buildkite/test_collector/tracer.rb
@@ -35,21 +35,30 @@ module Buildkite::TestCollector
         @children = []
       end
 
+      def duration
+        raise IncompleteSpan if end_at.nil?
+
+        end_at - start_at
+      end
+
       def as_hash
         {
           section: section,
           start_at: start_at,
           end_at: end_at,
-          duration: end_at - start_at,
+          duration: duration,
           detail: detail,
           children: children.map(&:as_hash),
         }
       end
+
+      class IncompleteSpan < StandardError; end
     end
 
-    def initialize
+    def initialize(min_seconds: nil)
       @top = Span.new(:top, MonotonicTime.call, nil, {})
       @stack = [@top]
+      @min_seconds = min_seconds
     end
 
     def enter(section, **detail)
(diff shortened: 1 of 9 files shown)


lib/buildkite/test_collector/tracer.rb
```diff
@@ -35,21 +35,30 @@ def initialize(section, start_at, end_at, detail)
         @children = []
       end
 
+      def duration
+        raise IncompleteSpan if end_at.nil?
```
niceking, line 39: I can't see this error being caught anywhere? If that's the case, I think this will might cause Rspec to fail with that "error outside of execution" error message and/or fail the job?

pda: Hmm yeah.

I think this is what Java would call an `IllegalStateException`, and is basically a runtime assertion that some internal usage hasn't gone wrong; calling `duration` before setting `end_at` makes no sense, and as far as we can tell nothing is or should be doing that.

My normal default is to crash with a clear error message when something “impossible” happens.

But perhaps I need to rethink that in test collectors where it's more important that the show goes on, even if that means the test collector can't behave sanely. 🤔

I'm pretty sure without this check, the old code would have done this instead:

```
NoMethodError: undefined method `-' for nil:NilClass
```

So I don't think it's a regression.

pda:
I wonder if configuring this as milliseconds (integer?) would be better than seconds (decimal/float).

I find much easier to express 1 second in milliseconds (1,000) than 1 millisecond in seconds (0.001?).
And I suspect people would generally want to configure this somewhere in the 1–1000ms range, not some whole number of seconds.

I think I'll change it.

pda:
I've tested this against a large build within Buildkite with `BUILDKITE_ANALYTICS_TRACE_MIN_MS=10`, and verified that I see spans over 10ms and none under 10ms 👌🏼

lib/buildkite/test_collector/tracer.rb
niceking: Oh right I see, you're setting the `end_at` right before the `duration` method is called - so the failure case would be if the call to Monotonic time library fails, which "should be" impossible. Is it worth putting in such a check for a dependant library? Not sure why the time call would fail? And seeing as this would have raised an exception in the past anyways, and we've never seen a case of that with the collector, maybe its fine to just rely on monotonic time doing its thing and not having an explicit check for it?

niceking: bump on this comment :)

niceking (approved):
I think raising the error slightly confuses me as it makes me think that this is a likely case that we should protect from, but it is as @pda pointed out not a regression and we've never had this happen in the while so its pretty harmless to leave in 🤷‍♀️ 

Excited to support this feature tho!! Big spans only!!!!

lib/buildkite/test_collector/tracer.rb
pda: Yeah I think it's fine; I don't think the monotonic time check can realistically return `nil`, if anything it might raise an error and crash the whole party.

So yeah I think it's fine like this.

PR #218, opened 2024-05-07
Run tests on a new queue

yob:
We're starting to test a new queue, and I'm looking for projects to experiment with. These new agents are quite different, but have docker and bash so I expect things should Just Work.

If there's any surprises, we can safely revert.

The new agents aren't authenticated to ECR, so I've switched to loading ruby docker images from the public ECR gallery. That's our preferred style these days anyway - agents that are authenticated to ECR will transparently use ECR as a cache. We don't need to explicitly use ECR images.

### DIFF
--- a/buildkite.yaml
+++ b/buildkite.yaml
@@ -1,3 +1,6 @@
+agents:
+  queue: hosted
+
 steps:
   - label: ":rspec: Tests :ruby: 3.3-rc"
     command:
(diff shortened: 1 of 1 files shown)


niceking:
Happy for you to experiment, although we don't do a huge number of builds on this pipeline 😅
Screenshot 2024-05-07 at 12 33 06 PM
I guess that means its low risk - but also might not provide a lot of validation, other than your existing branch? But I'm fine with that if you're fine with that 🤷‍♀️

niceking (approved):
Read your comment in slack about it not being high volume 😂

yob:
I agree this isn't going to provide a lot of ongoing signal, but even just opening the branch and seeing that yet another pipeline works on the new queue with no changes is great intel. Having come this far, may as well make the new queue the default and maybe we'll get a bit of extra data at some point ❤️

PR #220, opened 2024-05-14
Add configurable span filters.

catkins:
When using the collector, sometimes there can many uninteresting spans in the trace from things like selenium or other framework noise. This adds a seam to allow users to have control of filtering out uninteresting trace spans.

## Usage

```rb
ignore_selenium_requests = ->(span) do
  return true unless span[:section] == "http"

  !span.dig(:detail, :url) =~ /^http://selenium:4444//
end

Buildkite::TestCollector.configure(hook: :rspec)
Buildkite::TestCollector.span_filters << ignore_selenium_requests
```

This doesn't need to be a lambda, any `#call`-able will work, as long as it returns a truthy value from the callback when we want to retain the span.

Multiple `span_filters` can be provided, and as long they _all_ return a truthy value, then the span will be retained, this allows composing multiple filters together.

This would be a public interface to the library, so I'm happy to workshop the naming/semantics.

### DIFF
--- a/lib/buildkite/test_collector.rb
+++ b/lib/buildkite/test_collector.rb
@@ -39,12 +39,14 @@ module Buildkite
       attr_accessor :env
       attr_accessor :batch_size
       attr_accessor :trace_min_duration
+      attr_accessor :trace_ignore_span
     end
 
-    def self.configure(hook:, token: nil, url: nil, tracing_enabled: true, artifact_path: nil, env: {})
+    def self.configure(hook:, token: nil, url: nil, tracing_enabled: true, trace_ignore_span: nil, artifact_path: nil, env: {})
       self.api_token = (token || ENV["BUILDKITE_ANALYTICS_TOKEN"])&.strip
       self.url = url || DEFAULT_URL
       self.tracing_enabled = tracing_enabled
+      self.trace_ignore_span = trace_ignore_span || ->(span) { false }
       self.artifact_path = artifact_path
       self.env = env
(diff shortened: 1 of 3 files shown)


wooly:
My gut says that the use of the word 'filter' in the option is throwing me since I'm associating it with Array#Filter, which I think has the opposite semantics of this: 'keep the stuff for which the block returns true'. I know that it's inverted internally but it just feels ever so slightly off.

What if the option was just `span_filter` and it filtered out everything that returned false?

catkins:
Yeah that's what I started with too actually and I went back and forth a bit. 

Perhaps I'll make a more interesting filter that rejects a few other types of span to see how that would look.

catkins:
It would be great if there was a block form of `Buildkite::TestCollector.configure` similar to how rspec or vcr get setup

```rb
Buildkite::TestCollector.configure(hook: :rspec do |config|
  config.span_filter = ->(span) do
    # ... filtering things
  end
end

```

wooly:
Yeah that feels like a much more expressive API for sure.

catkins:
I shuffled things around slightly to be composable, and re-expressed the minimum duration filtering @pda added in #215 to use the new construct.

I also updated the PR description to show the new syntax.

wooly:
Looks neat!

zhming0 (approved):
I want this!! 🤩

catkins:
@zhming0 fancy taking over the PR? I haven't had my head in the TA space just at the moment, and it could be good for you to have a play with doing a collector release

PR #223, opened 2024-09-16
Handle http response errors during upoad.

zhming0:
part of PIE-2813

Fix uploader so it will throw a runtime error if backend does not return 2xx. 
This will result in a log message if any upload fails.

Note it will not fail a build if the upload fails.

### DIFF
--- a/lib/buildkite/test_collector/http_client.rb
+++ b/lib/buildkite/test_collector/http_client.rb
@@ -38,7 +38,13 @@ module Buildkite::TestCollector
 
       contact.body = compressed_body.string
 
-      http.request(contact)
+      response = http.request(contact)
+
+      if response.is_a?(Net::HTTPSuccess)
+        response
+      else
+        raise "HTTP Request Failed: #{response.code} #{response.message}"
+      end
     end
 
     def metadata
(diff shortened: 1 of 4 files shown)


wooly:
I'm not super keen on this, since it will cause people's builds to fail if the upload API is down. I always thought of the test collectors as a best-effort service.

What do you think about making this a configuration option so people could turn it on if they so wished?
How does this behaviour compare to the other collectors? Do they raise an issue if the data cannot be sent?

zhming0:
> I'm not super keen on this, since it will cause people's builds to fail if the upload API is down. I always thought of the test collectors as a best-effort service.

@wooly and @nprizal 

It won't fail people's build. Our uploader has[ logic capturing the error](https://github.com/buildkite/test-collector-ruby/blob/9bb438559c9cd0b7b838f06a0f3ac458e8fb3d92/lib/buildkite/test_collector/uploader.rb#L48-L51). 

It was just previously, the error reporting was never working, so we were completely in the dark about what happen in the collector.

KatieWright26 (approved):

gchan (approved):
lgtm, have we considered adding a tests to verify this new functionality? I assume you have manually tested this and viewed the new log message.

wooly:
@zhming0 ah yep I see it now, sorry for the noise!

zhming0:
@gchan I added some tests. 👍🏿

PR #226, opened 2024-11-03
Add execution_count as part of our upload.run_env

zhming0:
It's proven tricky to debug individual upload due to the fact that they are normally gzipped. Correlating gzipped content with other parts of our system is labor intensive. 

For example, finding how many executions do each upload has is a pretty common task but really non-trivial to do in our current backend. 

This PR add a execution_count fields on upload metadata `run_env`, this will allow us to answer the question above trivially.

### DIFF
--- a/lib/buildkite/test_collector/http_client.rb
+++ b/lib/buildkite/test_collector/http_client.rb
@@ -25,7 +25,9 @@ module Buildkite::TestCollector
       data_set = data.map(&:as_hash)
 
       body = {
-        run_env: Buildkite::TestCollector::CI.env,
+        run_env: Buildkite::TestCollector::CI.env.merge({
+          execution_count: data_set.size
+        }),
         format: "json",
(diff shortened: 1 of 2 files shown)


nprizal (approved):
Where did my review gone?

A thing to consider is sending the `execution_count` as a separate field instead of adding it to `run_env`. `run_env` to me, is more about metadata related to the environment where the test is running (hence the name). While `execution_count` is a metadata for the actual data that we send to TE.

My concern is around `run_env` becoming a dumping ground that will hard to manage. However, I understand that this the pattern that we've been doing so far, so I'm going to approve this PR.

zhming0:
This PR was introduced when we had little debugging ability for the upload process. Since then we have added a few new debugging tools, we probably no longer need this PR, as it's a bit too narrow to be useful. 

Thanks for reviewing this though @nprizal !

PR #228, opened 2024-12-12
Remove failed test links from the job log

malclocke:
These links have been superceded by the "Test Digest" tab on the build page.

PR #230, opened 2024-12-16
Remove call to deleted formatter

malclocke:
This formatter was removed in
https://github.com/buildkite/test-collector-ruby/pull/228 but this reference was accidentally left behind.


gchan (approved):

gchan (approved):
:+1: great mitigation and comment in the spec helper file!

PR #232, opened 2025-02-03
Development dependencies bumped for Ruby 3.4

pda:
Local development / testing was broken on latest Ruby 3.4 for a couple of reasons:

- `Gemfile.lock` specified `activesupport (7.0.7.2)` which does `require "mutex_m"`, and that has moved from `stdlib` to a bundled gem. Newer `activesupport` declares `mutex_m` as a dependency.
- Ruby 3.4 is stricter about `Hash.new(…)` being passed unrecognised keyword arguments, which we were mistakenly doing in `spec/support/rspec_example_trace_helpers.rb`.

I've done a `bundle update` which moved the `Gemfile.lock` (development/testing only) dependency to `activesupport (8.0.1)`. This doesn't affect the runtime dependencies; we're still aiming to support older Ruby / Rails etc.


gchan:
It appears that requiring securerandom 0.4.1 is breaking the Ruby 2.7 and 3.0 builds

`securerandom-0.4.1 requires ruby version >= 3.1.0`

securerandom 0.3.2 works in Ruby 2.6 and above. Maybe we could give that a go?

gchan (approved):

PR #237, opened 2025-02-04
Create new pipeline for releasing to rubygems

yob:
This adds a new pipeline.yml specifically for pushing gem releases to rubygems.org. We use the `rubygems-oidc` plugin to swap a Buildkite OIDC token for a temporary rubygems.org token that has permission to push this gem - that exchange process will only work from a special Buildkite pipeline that is triggered by a tag push to the repository.

I've updated the README with the new process. It's largely the same, except now any staff member with access to the release pipeline can approve the release. They no longer need to be added as an owner on [the gem](https://rubygems.org/gems/buildkite-test_collector), or need to have access to any shared credentials for our rubygems.org user.

### DIFF
--- a/.buildkite/pipeline.release.yml
+++ b/.buildkite/pipeline.release.yml
@@ -1,0 +1,18 @@
+agents:
+  queue: hosted
+
+steps:
+  - block: "OK to release?"
+
+  - command: ".buildkite/steps/release-gem"
+    label: ":rubygems:"
+    if: build.tag != null
+    plugins:
+      - rubygems-oidc#v0.2.0:
+          role: "rg_oidc_akr_fy1x4px4yjwd1rdhkkda"
+      - docker#v5.12.0:
+          image: "ruby:3.4"
+          environment:
+            - GEM_HOST_API_KEY
+            - BUILDKITE_TAG
+
--- a/.buildkite/steps/release-gem
+++ b/.buildkite/steps/release-gem
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+set -euo pipefail
+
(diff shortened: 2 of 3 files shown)


.buildkite/steps/release-gem
```diff
@@ -0,0 +1,58 @@
...
+if [ $(curl -s -o /dev/null -w "%{http_code}" https://rubygems.org/api/v2/rubygems/pdf-reader/versions/${VERSION}.json) == "200" ]; then
+  echo "Gem version ${VERSION} already found on rubygems, skipping release"
+  exit 1
+fi
+
+echo "--- publish gem"
+
+# Skip while testing
+# gem push "${GEM_FILENAME}"
+echo "pretend push ${GEM_FILENAME}"
```
yob, line 58: After this has been reviewed and before it merges, we should change this to plain `gem push .... `. I had it like this for testing

yob:
I'm not thrilled with the amount of bespoke bash I had to right for this. It'd be nice to ... not have it. Maybe later we can extract the pattern into a "rubygems-release" plugin that pairs well with the "rubygems-oidc" plugin or something....

.buildkite/pipeline.release.yml
```diff
@@ -0,0 +1,18 @@
+agents:
+  queue: hosted
+
+steps:
+  - block: "OK to release?"
+
+  - command: ".buildkite/steps/release-gem"
+    label: ":rubygems:"
+    if: build.tag != null
```
yob, line 9: https://buildkite.com/docs/pipelines/configure/conditionals#example-expressions

pda (approved):

PR #239, opened 2025-02-06
API endpoint configurable with `BUILDKITE_ANALYTICS_ENDPOINT`

pda:
Make the API endpoint configurable from the `BUILDKITE_ANALYTICS_ENDPOINT` environment variable.

It defaults to `https://analytics-api.buildkite.com/v1/uploads`, was previously only configurable by Ruby code:

```ruby
Buildkite::TestCollector.configure(url: "http://localhost:1234/example", …)
```

But it can be useful to inject different endpoints without changing the code, for testing etc.


niceking (approved):

niceking:
Change looks fine, looks like a test needs to be updated tho

PR #240, opened 2025-02-19
`Buildkite::TestCollector.tag_execution(key, value)`

pda:
Draft support for per-execution tagging.

- [x] Decide if `Buildkite::TestCollector.tag_execution(key, value)` is an okay API for this.
- [x] Add some test coverage.

I don't love adding yet another `Thread.current[:_buildkite…]` fiber-local (global-ish) name, but I don't see a really good place to put these tags otherwise, and really it's an inconsequential implementation detail, compared to the public API in front of it. I don't think they really belong in `Buildkite::TestCollector::Tracer` (we're tagging the execution, not the potentially-nested traces therein) and adding them there just adds complexity.

### DIFF
--- a/lib/buildkite/test_collector.rb
+++ b/lib/buildkite/test_collector.rb
@@ -77,6 +77,18 @@ module Buildkite
       tracer&.leave
     end
 
+    # Set a key=value tag on the current test execution.
+    def self.tag_execution(key, value)
+      tags = Thread.current[:_buildkite_tags]
+      raise "_buildkite_tags not available" unless tags
+
+      unless key.is_a?(String) && value.is_a?(String)
+        raise ArgumentError, "tag key and value expected string"
+      end
+
+      tags[key] = value
+    end
+
     def self.enable_tracing!
       return unless self.tracing_enabled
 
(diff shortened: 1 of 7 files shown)


lib/buildkite/test_collector/minitest_plugin.rb
```diff
@@ -14,18 +14,29 @@ def before_setup
...
 
   def after_teardown
     tracer = Thread.current[:_buildkite_tracer]
+    tags = Thread.current[:_buildkite_tags]
+
+    Thread.current[:_buildkite_tracer] = nil
+    Thread.current[:_buildkite_tags] = nil
+
     if !tracer.nil?
-      Thread.current[:_buildkite_tracer] = nil
```
zhming0, line 25: I wonder if there was a reason that this is put behind this if nil condition?

zhming0 (approved):

lib/buildkite/test_collector.rb
```diff
@@ -77,6 +77,18 @@ def self.annotate(content)
       tracer&.leave
     end
 
+    # Set a key=value tag on the current test execution.
+    def self.tag_execution(key, value)
+      tags = Thread.current[:_buildkite_tags]
+      raise "_buildkite_tags not available" unless tags
+
+      unless key.is_a?(String) && value.is_a?(String)
```
gchan, line 85: Should we consider supporting String like classes like numbers and booleans?

Should we be more strict about sanitisation or should we handle that server side?

pda: Unsure, so I started strict and we can expand it to be more permissive without breaking anything for anyone.
(Can't do the other direction)

gchan (approved):
:shipit:

PR #243, opened 2025-02-26
PIE-3345: fix ECR rate limit problem on CI

zhming0:
Our CI has been pretty unreliable recently. This PR fixes that. 
This is also serving as an example to how this is done.

Credits to @lucaswilric for working out the IAM puzzle! 

Completes PIE-3345

### DIFF
--- a/buildkite.yaml
+++ b/buildkite.yaml
@@ -11,10 +11,10 @@ steps:
       - aws-assume-role-with-web-identity#v1.2.0:
           role-arn: "${OIDC_ASSUME_ROLE_ARN:-arn:aws:iam::445615400570:role/pipeline-buildkite-test-collector-ruby-build}"
           session-tags:
-          - organization_slug
-          - organization_id
-          - pipeline_slug
-          - build_branch
+            - organization_slug
+            - organization_id
+            - pipeline_slug
+            - build_branch
 
       - ecr#v2.9.0:
           login: true
--- a/buildkite.yaml
+++ b/buildkite.yaml
@@ -18,7 +18,7 @@ steps:
 
       - ecr#v2.9.0:
           login: true
-          account_ids: "445615400570"
+          account_ids: "public.ecr.aws"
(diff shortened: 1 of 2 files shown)


buildkite.yaml
```diff
@@ -8,6 +8,18 @@ steps:
       - "bundle"
       - "bundle exec rake"
     plugins:
+      - aws-assume-role-with-web-identity#v1.2.0:
+          role-arn: "${OIDC_ASSUME_ROLE_ARN:-arn:aws:iam::445615400570:role/pipeline-buildkite-test-collector-ruby-build}"
+          session-tags:
+          - organization_slug
+          - organization_id
+          - pipeline_slug
+          - build_branch
```
pda, line 17: YAML is frustrating in its ambiguous indentation rules.
In this case, the array nested under `plugins` a few lines up has the hyphens indented past their container, which is my preferred style. Whereas this one has the hyphens a the same level as their container. My preference doesn't really matter, but we should probably be consistent within a single file.

```suggestion
            - organization_slug
            - organization_id
            - pipeline_slug
            - build_branch
```

buildkite.yaml
```diff
@@ -27,6 +39,17 @@ steps:
           - "bundle"
           - "bundle exec rake"
         plugins:
+          - aws-assume-role-with-web-identity#v1.2.0:
+              role-arn: "${OIDC_ASSUME_ROLE_ARN:-arn:aws:iam::445615400570:role/pipeline-buildkite-test-collector-ruby-build}"
+              session-tags:
+              - organization_slug
+              - organization_id
+              - pipeline_slug
+              - build_branch
```
pda, line 48:
```suggestion
                - organization_slug
                - organization_id
                - pipeline_slug
                - build_branch
```

pda (approved):
Fantastic 🤩

This has been really bugging me, but I had done the lazy thing and clicked retry about a hundred time 🙈

PR #245, opened 2025-05-28
Initial test collector support for Cucumber

tommeier:
Very rough, untested - to be tested next.

image


lib/buildkite/test_collector/library_hooks/cucumber.rb
```diff
@@ -12,6 +12,8 @@
 # -------------------------------------------------
 
 Before do |scenario|
+  puts scenario.class
+
```
nprizal, line 16:
```suggestion
```

spec/test_collector/cucumber_plugin/trace_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  describe '#as_hash' do
+    it 'removes invalid UTF-8 characters from top level values' do
+      failure_reason = trace.as_hash[:failure_reason]
+
+      expect(failure_reason).to be_valid_encoding
+    end
+
+    it 'removes invalid UTF-8 characters from nested values' do
```
tommeier, line 44: 👏

wooly:
Need to double check which versions of ruby we need to support with this, as cucumber 9 requires ruby 2.7 and above.

wooly:
Have tested this locally and it is uploading executions to TE as it should.

buildkite-test_collector.gemspec
```diff
@@ -27,4 +27,8 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "activesupport", ">= 4.2"
   spec.add_development_dependency "rspec-core", '~> 3.10'
   spec.add_development_dependency "rspec-expectations", '~> 3.10'
+
+  if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7')
```
malclocke, line 31: **question** is this needed because of the legacy build steps in `buildkite.yaml`?

malclocke (approved):

buildkite-test_collector.gemspec
malclocke: Yeh, but it's not immediately obvious to me why this check is in **this** file.

The committed `Gemfile.lock`, which was presumably generated by running `bundle` with a version >= 2.7, has cucumber in it.

This file (`buildkite-test_collector.gemspec`) is only used in development (which will presumably be done with something > 2.7) and test. The [build steps for legacy versions](https://github.com/buildkite/test-collector-ruby/blob/73763fcadea40fad368b0e8319865eb3e1b3c995/buildkite.yaml#L38-L40) run bundle, which will overwrite `Gemfile.lock` with a version that doesn't include cucmber when the ruby version is < 2.7.

So I **think** it's only the legacy build steps that need this. I'd suggest we add a comment here because it's not obvious IMO.

```suggestion
  # When running the legacy CI builds against versions of Ruby pre 2.7 we cannot include cucumber 9 as it's not supported.
  if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7')
```

wooly: Ah! Now I see what you're saying, and yeah I agree that makes it much clearer. Have added ☺️

PR #247, opened 2025-06-05
Set BUILDKITE_ANALYTICS_TOKEN on CI

malclocke:


buildkite.yaml
```diff
@@ -23,6 +26,9 @@ steps:
           region: "us-east-1"
       - docker#v3.7.0:
           image: "public.ecr.aws/docker/library/ruby:{{matrix}}"
+          propograte-environment: true
```
pda, line 29:
```suggestion
          propagate-environment: true
```

Chris-tian30 (approved):

features/support/env.rb
```diff
@@ -1,4 +1,4 @@
 require 'buildkite/test_collector'
 Buildkite::TestCollector.configure(
-  hook: :cucumber, url: "http://analytics-api.buildkite.localhost/v1/uploads"
+  hook: :cucumber, url: "https://analytics-api.buildkite.com/v1/uploads"
```
nprizal, line 3:
```suggestion
  hook: :cucumber
```
Will this work? I believe it will use the prod URL by default if the `url` is not specified.

PR #260, opened 2026-07-13
Fix file_name truncation in RSpec/Cucumber FILE_PATH_REGEX

nprizal:
We want to tag `test.selector.primary` for test results uploaded by test collectors. We can't do that currently for the Ruby collectors (rspec, cucumber) because `file_name` extraction has an edge case where it can silently truncate valid paths.

The problem is `FILE_PATH_REGEX` (`/^(.*?\.(rb|feature))/`), a non-greedy match up to the *first* `.rb`/`.feature` occurrence in the string. Nothing stops a path from having one of those extensions earlier than the real file extension, e.g. `spec/archive.rb.d/widget_spec.rb`, and if that happens, it truncates to `spec/archive.rb`. It's an unusual naming pattern and unlikely to show up in practice, but `ta-ingestion` can't rely on `file_name` while it's even possible, so it's blocking `test.selector.primary` inference for this collector.

`FILE_PATH_REGEX` was reused against three differently-shaped strings, so instead of patching the regex, I replaced it with a targeted suffix-strip for each shape. This is a more robust extraction strategy in general, not just a fix for the one example above:

- `example.id`, e.g. `./spec/foo_spec.rb[1:2]` → strip the trailing `[...]` index suffix
- `example.location`, e.g. `./spec/foo_spec.rb:42` → strip the trailing `:<lineno>`
- `shared_example_call_location` (a backtrace line), e.g. `` ./spec/foo_spec.rb:12:in `block (2 levels) in <top (required)>' `` → strip the trailing `:<lineno>:in ...` frame label

The existing mismatch-detection/fallback branch for shared examples (`it_behaves_like`) is untouched, it still compares the id and location paths and falls back to `shared_example_call_location` when they diverge.

Applied the same fix to the Cucumber plugin, which had its own copy of the same regex.

Also, while working on this, I found `http_client_spec.rb` had a fake example with unrealistic `id`/`location` values that happened to both fail to match the old regex (so they were coincidentally equal), which masked a `NoMethodError` from missing `metadata` once the extraction got fixed. Updated it with realistic values.

## Test plan

- [x] Regression test covering a nested path like `spec/archive.rb.d/widget_spec.rb`
- [x] Regression test covering a shared example (`it_behaves_like`) proving the actual spec file is returned, not the shared-examples library file
- [x] Same fix applied to Cucumber plugin, with equivalent regression tests
- [x] Full test suite passes (`bundle exec rspec`, 71 examples, 0 failures)

### DIFF
--- a/lib/buildkite/test_collector/cucumber_plugin/trace.rb
+++ b/lib/buildkite/test_collector/cucumber_plugin/trace.rb
@@ -5,7 +5,9 @@ module Buildkite::TestCollector::CucumberPlugin
     attr_accessor :scenario, :failure_reason, :failure_expanded
     attr_reader :history, :tags, :location_prefix
 
-    FILE_PATH_REGEX = /^(.*?\.(rb|feature))/
+    # scenario.location looks like "./features/foo.feature:12" - strip the
+    # trailing line number to get the file path.
+    LOCATION_LINE_SUFFIX_REGEX = /:\d+\z/
 
     def initialize(scenario, history:, failure_reason: nil, failure_expanded: [], tags: nil, location_prefix: nil)
       @scenario         = scenario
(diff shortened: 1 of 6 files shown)


nprizal:
Defer this for now as this is avery rare case. We can come back to this if this is becoming a real problem.

PR #261, opened 2026-07-23
PoC: OpenTelemetry traces for RSpec test executions (TE-6490)

nprizal:
## Summary

This developer-facing PoC adds opt-in OpenTelemetry export for RSpec test executions so we can dogfood test-scoped traces with Buildkite Test Engine.

Each sampled example creates an independently rooted `test.execution` span. Synchronous HTTP, SQL, Redis, and other instrumented work becomes its descendants. The normal Test Engine execution upload remains authoritative and independent: OpenTelemetry configuration and delivery failures warn or drop telemetry without failing tests or blocking the upload.

This is intentionally a v3 experiment. It requires Ruby 3.3+ and packages the tested OpenTelemetry SDK, OTLP exporter, and instrumentation set as core dependencies.

## Implemented contracts

- A shared opaque ID correlates each sampled root with its execution upload; it is not copied to child spans.
- The validated raw run key is sent as both `Buildkite-Test-Run-Key` and `buildkite.test.run.key`. The authenticated receiver—not the collector—derives the canonical suite-scoped run UUID.
- A canonical `BUILDKITE_JOB_ID` from the Buildkite process environment is sent consistently in the job header and resource attributes. Legacy or configured overrides cannot manufacture typed job identity.
- Valid Agent `TRACEPARENT` / `TRACESTATE` becomes one SpanLink created with the root; it never becomes the parent.
- Buildkite resource metadata is merged only into the Buildkite export copy. Customer provider resources and exporter data are not mutated.
- Flush and shutdown affect only the Buildkite processor. Existing customer telemetry lifecycle remains customer-owned.
- OTLP activation is RSpec-only and independent of the collector's legacy `tracing_enabled` option.

## Dogfooding scope

This PR is evidence gathering, not a production support commitment. It deliberately uses broad `use_all` / `install_all` instrumentation and SDK-default batching. Static skips, production sampling and limits, an instrumentation allowlist, other test frameworks, sensitive-data policy, operating cost, and delivery guarantees remain decisions for the dogfooding exercise.

The two review documents are:

- [Architecture](https://github.com/buildkite/test-collector-ruby/blob/nprizal/te-6490-otel-ruby-spans-poc/docs/opentelemetry-architecture.md): implemented flow, identity contracts, metadata, ownership, and failure boundaries.
[...]

PR #262, opened 2026-08-06
Generate an external_id for each RSpec execution

nprizal:
We want to start ingesting OTel spans, eventually replacing our in-house span tracing. To join an execution with its OTel spans, each execution needs an opaque, time-sortable ID we can send to the Upload API.

I generated a UUIDv7 per RSpec execution and included it in the upload payload as `external_id`. UUIDv7 gives time ordering and storage locality once it's persisted (matches the approach from the OTel PoC, #261).

RSpec only for now. We're only dogfooding OTel from the RSpec hook, so Minitest and Cucumber don't need this yet.

`uuid_v7` needs Ruby 3.3+, but this gem still supports Ruby 2.3+, so `Buildkite::TestCollector::UUID.v7` falls back to UUIDv4 on older Rubies. Once we bump the required Ruby version for the OTel work, we can drop the fallback and call `SecureRandom.uuid_v7` directly.

### DIFF
--- a/lib/buildkite/test_collector/library_hooks/rspec.rb
+++ b/lib/buildkite/test_collector/library_hooks/rspec.rb
@@ -37,7 +37,8 @@ RSpec.configure do |config|
         example,
         history: tracer.history,
         tags: tags,
-        location_prefix: Buildkite::TestCollector.location_prefix
+        location_prefix: Buildkite::TestCollector.location_prefix,
+        external_id: Buildkite::TestCollector::UUID.v7,
       )
 
       Buildkite::TestCollector.uploader.traces[example.id] = trace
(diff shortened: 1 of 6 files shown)


mattensor (approved):
The internal UUID generation looks great as a fallback. I remember discussing allowing customers to provide their own external ID, with the collector generating one when none is provided. As a follow-up, it would be useful to expose a supported customer configuration option. likely a per-execution callback + document the external ID contract: 1–64 printable, non-space ASCII characters (^[!-~]{1,64}$) 🙌

nprizal:
> I remember discussing allowing customers to provide their own external ID, with the collector generating one when none is provided.

My understanding for this is for executions uploaded outside of test collector. I think for starter we should use test collector generated external_id, and allow this to be provided by customer if anyone asked for that. I don't think it is something we need to allow on our first release.

